PR #160, opened 2020-09-13
Add route tests and reorganize routes

zaucker:
Parameters might not all be correct yet. Will be fixed in next PRs together with functional tests for Agrammon::Web::Service.


dev/META6.json
```diff
@@ -11,12 +11,15 @@
...
     "Cro::HTTP",
     "Cro::HTTP::Session::Pg",
+    "Cro::HTTP::Test",
     "DB::Pg",
     "Digest::SHA1::Native",
     "OO::Monitors",
     "Shell::Command",
     "Spreadsheet::XLSX",
     "Text::CSV",
+    "Test",
```
jnthn, line 21: Hm, `Test` is provided in core, so it isn't usually mentioned in the `depends`.

dev/META6.json
```diff
@@ -11,12 +11,15 @@
   "depends": [
     "Cro::HTTP",
     "Cro::HTTP::Session::Pg",
+    "Cro::HTTP::Test",
```
jnthn, line 14: Not a big issue, but strictly belongs in `test-depends`.

lib/Agrammon/Web/Routes.pm6
```diff
@@ -5,100 +5,67 @@ use Cro::HTTP::Router;
 use Agrammon::Web::Service;
 use Agrammon::Web::SessionUser;
 
-
 sub routes(Agrammon::Web::Service $ws) is export {
+    subset LoggedIn of Agrammon::Web::SessionUser where .logged-in;
```
jnthn, line 9: It's clearer to put this entirely outside of `routes`, at the top level, given it's package-scoped.

lib/Agrammon/Web/Routes.pm6
```diff
@@ -5,100 +5,67 @@ use Cro::HTTP::Router;
...
+
+    # Fix for Agrammon
+    # my $root = $app ~~ Easii::Application::Development ?? 'frontend/source-output' !! 'qx-build';
```
jnthn, line 12: Maybe don't mention Easii? :-) Also, I'm curious if `$root` is over-used in the static content? It seems primarily relevant for the scripts, but maybe not the HTML?

lib/Agrammon/Web/Routes.pm6
```diff
@@ -213,24 +172,82 @@ sub routes(Agrammon::Web::Service $ws) is export {
...
-            request-body -> (:$dataset!) {
-                my $data = $ws.get-input($user, $dataset);
+sub user-routes(Agrammon::Web::Service $ws) {
+    route {
+
+        # implement/test
+        post -> LoggedIn $user, 'change_password' {
+            request-body -> (:$old-password!, :$new-password!) {
+                my $data = $ws.change-password($old-password, $new-password);
+                content 'application/json', $data;
```
jnthn, line 190: Looking at this made me a bit curious: how does `$ws` know which user it is to change the password for? Should we be passing something to it? Same for various other operations.

lib/Agrammon/Web/Service.pm6
```diff
@@ -84,8 +119,15 @@ class Agrammon::Web::Service {
...
+        ...
+        $user.change-password($oldPassword, $newPassword);
+        return $user;
+    }
+
+    method reset-password(Agrammon::Web::SessionUser $user, Str $email, Str $password, Str $key) {
+        ...
+        $user.reset-password($email, $password);
         return $user;
     }
```
jnthn, line 132: Here I see it *does* expect to be passed the user. So I guess the route code needs updating.

t/routes-application.t
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+done-testing;
+
+=begin pod
+
+=COPYRIGHT Copyright (c) 2020 by OETIKER+PARTNER AG. All rights reserved.
+
+=AUTHOR S<Fritz WorthZauckerington E<lt>fritz.zaucker@oetiker.chE<gt>>
```
jnthn, line 191: Who? :rofl:

t/routes-application.t
```diff
@@ -0,0 +1,195 @@
...
+}
+
+# my $fake-user = %( :id(1), :username('foor@bar.ch'), );
+
+# my $fake-tag-a = %( :id(42), :name('TagA'), :user($fake-user));
+# my $fake-tag-b = %( :id(43), :name('TagB'), :user($fake-user));
+
+# my $fake-dataset-a = %( :id(42), :name('DatasetA'), :user($fake-user));
+# my $fake-dataset-b = %( :id(43), :name('DatasetB'), :user($fake-user));
+
```
jnthn, line 25: Maybe not applicable in this test?

lib/Agrammon/Web/Service.pm6
jnthn: Ah, now I see the "Parameters might not all be correct yet" mention at the top of the PR, maybe it's about this.

dev/META6.json
zaucker: Removed

dev/META6.json
zaucker: Moved together with the other test related modules.

lib/Agrammon/Web/Routes.pm6
zaucker: Moved

lib/Agrammon/Web/Routes.pm6
zaucker: Removed Easii ... would like to keep `$root` as is until I have the frontend build (for both local and deployed) figured out.

t/routes-application.t
zaucker: OMG ... fixed.

t/routes-application.t
zaucker: Removed

lib/Agrammon/Web/Routes.pm6
zaucker: In `Agrammon::Web::SessionUser` I do set the `username` attribute to the login user.
What if I also set the `id` attribute as well? Then I can use the SessionUser everywhere I need the logged-in user.

Or is this a bad design?

lib/Agrammon/Web/Service.pm6
zaucker: See above.

lib/Agrammon/Web/Routes.pm6
```diff
@@ -185,8 +185,8 @@ sub user-routes(Agrammon::Web::Service $ws) {
 
         # implement/test
         post -> LoggedIn $user, 'change_password' {
-            request-body -> (:$oldPassword!, :$newPassword!) {
-                my $data = $ws.change-password($oldPassword, $newPassword);
+            request-body -> (:$old-password!, :$new-password!) {
```
jnthn, line 188: You're fine with the JSON object we get having to have the hyphens in them? (If not, you can do `:oldPassword($old-password)` to use a Raku-style variable name but extract it from another key in the JSON).

lib/Agrammon/Web/Routes.pm6
jnthn: No, that seems sensible enough to me. I mean, carrying around huge amounts of user data in there can make the session expensive to save/load (though we can also lazily populate it), but I don't think one extra ID attribute is a problem at all.

lib/Agrammon/Web/Routes.pm6
zaucker: I guess I overdid the Rakuization :-)

lib/Agrammon/Web/Routes.pm6
zaucker: As it is currently implemented, upon authentication **all** user info is loaded into the SessionUser object.
As it is not all that much info and we don't have large numbers of users, I think we can leave it like that for the moment.

lib/Agrammon/Web/Service.pm6
zaucker: > Ah, now I see the "Parameters might not all be correct yet" mention at the top of the PR, maybe it's about this.

Actually I meant that I have to bring the parameters the frontend sends in sync with what the backend expects. This is probably correct for most routes, but I will have to verify that and in some cases the parameters might be a bit strangely named.

However, I found it easier to make the backend somewhat consistent and add route tests for it first, then in the next PR add the missing webservice calls (with tests) and then in the third step bring the frontend in sync.

lib/Agrammon/Web/Routes.pm6
zaucker: With that setup, method calls on the SessionUser object will automagically work on the right database entry.

jnthn (approved):

lib/Agrammon/Web/Routes.pm6
```diff
@@ -287,27 +302,24 @@ sub routes(Agrammon::Web::Service $ws) is export {
...
             }
         }
 
         # test/implement
-        post -> LoggedIn $user, 'order_instances', $name {
-            ...
-            request-body -> (:$name!) {
-                my $data = $ws.order-instances($user, $name);
+        post -> LoggedIn $user, 'order_instances' {
+            request-body -> (:@instances, :$dataset-name!) {
```
jnthn, line 321: Probably `:datasetName($dataset-name)`?

lib/Agrammon/Web/Routes.pm6
```diff
@@ -287,27 +302,24 @@ sub routes(Agrammon::Web::Service $ws) is export {
 
         # test/implement
         post -> LoggedIn $user, 'store_branch_data' {
-                ...
-            request-body -> (:$name!) {
-                my $data = $ws.store-branch-data($user, $name);
+            request-body -> (:%data!, :$dataset-name!) {
```
jnthn, line 305: `:datasetName($dataset-name)`

PR #163, opened 2020-09-16
Webservices and tests

zaucker:
This wires up all required routes, webservices, and backend modules. 

Once this is merged, I will verify all functionalities, adapt frontend/backend parameters, and add missing backend functionalities in separate PRs for each route.


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -27,13 +27,75 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+            my @d = $ds.array;
             $!id = @d[0];
             $!mod-date = @d[1];
         }
         return self;
     }
 
+    method rename($new) {
+        self.with-db: -> $db {
+            my $ret = $db.query(q:to/DATASET/, $new, $!name, $!user.id);
```
jnthn, line 39: `$ret` is unused

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -27,13 +27,75 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+    method !clone($new) {
+        my $dataset = Agrammon::DB::Dataset.new(:name($new));
+        warn "clone($new) not yet implemented";
+        return $dataset;
+    }
+
+    method submit($email) {
+        my $new-dataset = self!clone("Clone of $!name");
+
+        warn "Sending mail for submit($email) not yet implemented";
```
jnthn, line 58: Sending email is probably not best done in a `DB::Dataset` class anyway. :-)

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -27,13 +27,75 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+        }
+        return self;
+    }
+
+    method remove-tag($tagName) {
+        self.with-db: -> $db {
+            my $tag = Agrammon::DB::Tag.new( :name($tagName), :user($!user)).lookup;
+            my $ret = $db.query(q:to/DATASET/, $tag.id, $!id);
+                DELETE FROM tagds
+                 WHERE tagds_tag = $1 AND tagds_dataset = $2
```
jnthn, line 93: Does this need to check `$tag` lookup was successful and handle it in some good way?

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -27,13 +27,75 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+               AND dataset_name = $2
+            DATASET
+            $!id = $results.value;
+        }
+        return self;
+    }
+
+    method set-tag($tagName) {
+        self.with-db: -> $db {
+            my $tag = Agrammon::DB::Tag.new( :name($tagName), :user($!user)).lookup;
```
jnthn, line 78: Similar question here.

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -49,58 +111,120 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+    method store-input-comment($var-name, $comment) {
+        my $instance;
+
+        my $var = $var-name;
+        if $var ~~ s/\[(.+)\]/[]/ {
+            $instance = $0;
+        }
+
+        return $instance ?? self!store-instance-variable-comment($var, $instance, $comment)
+                         !! self!store-variable-comment($var, $comment);
```
jnthn, line 225: Is it correct that this passes `$var` down without removing the instance part?

lib/Agrammon/DB/Tag.pm6
```diff
@@ -1,22 +1,59 @@
...
+            TAG
 
             $!id = $results.value;
         }
         return self;
     }
 
+    method rename($new) {
+        self.with-db: -> $db {
+            my $ret = $db.query(q:to/TAG/, $new, $!name, $!user.id);
```
jnthn, line 25: Unused `$ret`

lib/Agrammon/DB/Tag.pm6
```diff
@@ -1,22 +1,59 @@
...
+                RETURNING tag_name
+            TAG
+            $!name = $new;
+        }
+        return self;
+    }
+
+   method delete {
+        self.with-db: -> $db {
+            my $ret = $db.query(q:to/TAG/, $!id);
```
jnthn, line 37: Unused `$ret`

lib/Agrammon/Test.pm6
```diff
@@ -0,0 +1,43 @@
+
+## THIS FILE TO BE REMOVED AS SOON AS Cro::HTTP::Test supplies these
```
jnthn, line 2: Hmm, I thought the PR was merged for those...

lib/Agrammon/Web/Routes.pm6
```diff
@@ -72,27 +72,30 @@ sub dataset-routes(Agrammon::Web::Service $ws) {
...
+                content 'application/json', $data;
+            }
         }
 
         post -> LoggedIn $user, 'send_datasets' {
-            ...
-            my $data = $ws.send-datasets($user);
-            content 'application/json', $data;
+            request-body -> (:@datasets!) {
+                dd 'datasets=' ~ @datasets;
```
jnthn, line 83: Debug leftover

lib/Agrammon/Test.pm6
jnthn: That said, I don't think you're actually using these helpers anyway?

lib/Agrammon/DB/Dataset.pm6
zaucker: Removed

lib/Agrammon/DB/Dataset.pm6
zaucker: True.

lib/Agrammon/DB/Tag.pm6
zaucker: Removed

lib/Agrammon/DB/Tag.pm6
zaucker: Removed for the moment. I guess when checking the frontend interaction, I might actually want to return the number of tags deleted or something like that.

lib/Agrammon/Test.pm6
zaucker: Yes, I think I committed this accidentally. Removed.

lib/Agrammon/Web/Routes.pm6
zaucker: Removed

lib/Agrammon/DB/Dataset.pm6
zaucker: We do remove the instance name (with the regex substitution). The DB stores the variable name with the empty `[]`.

lib/Agrammon/DB/Dataset.pm6
zaucker: Added a check (should not be necessary, as the frontend will only be able to request a valid tag, at least as long as the same user doesn't work on the same dataset from two different browsers).
A simple `if $tag.id` should do, as nothing bad happens if the tag was already removed.

lib/Agrammon/DB/Dataset.pm6
zaucker: Same as above, the user will at least see of the new tag is there. Will think about it a bit more when testing things together with the frontend.

lib/Agrammon/Web/Service.pm6
```diff
@@ -38,29 +38,40 @@ class Agrammon::Web::Service {
...
+    method delete-datasets(Agrammon::Web::SessionUser $user, @datasets) {
+        return @datasets.elems;
+        # return Agrammon::DB::Datasets.new(:$user, :$version).load.list;
+    }
+
+    # return list of datasets as expected by Web GUI
+    method send-datasets(Agrammon::Web::SessionUser $user, @datasets) {
+        return @datasets.elems;
+#        return Agrammon::DB::Datasets.new(:$user, :$version).load.list;
+    }
```
jnthn, line 50: Left as a reminder, or leftover commented code to remove?

jnthn (approved):

lib/Agrammon/Web/Service.pm6
zaucker: One of the things to now get into sync between frontend and backend.

PR #164, opened 2020-09-18
Fix store_data route

zaucker:
Adapt and cleanup frontend/backend interface.

### DIFF
--- a/lib/Agrammon/Web/Routes.pm6
+++ b/lib/Agrammon/Web/Routes.pm6
@@ -16,6 +16,15 @@ sub routes(Agrammon::Web::Service $ws) is export {
     my $schema = 'share/agrammon.openapi';
     my $root = '';
     route {
+        # before {
+        #     # Consume and re-instate request.
+        #     my $blob = await request.body-blob;
+        #     request.set-body($blob);
+        #     # Dump.
+        #     my $req = ~request;
+        #     try $req ~= $blob.decode('utf-8');
+        #     note $req;
+        # }
         include static-content($root);
         include api-routes($schema, $ws);
         include user-routes($ws);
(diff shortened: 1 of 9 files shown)


lib/Agrammon/Web/Routes.pm6
```diff
@@ -272,10 +281,14 @@ sub application-routes(Agrammon::Web::Service $ws) {
...
-        # test
+        # implement branches
         post -> LoggedIn $user, 'store_data' {
             request-body -> %data {
-                my $ret = $ws.store-data($user, %data);
+                my $ret = $ws.store-data(
+                    $user,
+                    :dataset(%data<dataset_name>), :var(%data<data_var>), :value(%data<data_val>),
+                    :branches(%data<branches>), :options(%data<options>)
+                );
```
jnthn, line 291: Did you consider unpacking these in `request-body`, e.g. like:

```
request-body (Str :dataset_name($dataset)!, Str :data_var($var)!) { # and the others
    my $ret = $ws.store-data($user, :$dataset, :$var); # and the others
}
```

This means that the data will be validated and a 400 issued if it is mis-typed or mis-named (for example, an array isn't sent for `options` or `branches`). Whereas now it'll type-check fail on the call to `store-data` and that'll be a 500.

t/webservice.t
```diff
@@ -158,19 +158,15 @@ transactionally {
...
-        %data = %(
-            :dataset_name('MyNewTestDataset'),
-            :data_var('Livestock::DairyCow[MK]::Excretion::dairy_cows'),
-            :data_val(42)
-        );
-        ok $ws.store-data($user, %data), "Instance input stored";
+        my $dataset = 'MyNewTestDataset';
+        my $var = 'PlantProduction::AgriculturalArea::agricultural_area';
+        my $value = 42;
+        ok $ws.store-data($user,  :$dataset, :$var, :$value), "Simple input stored";
```
jnthn, line 164: This form would not be callable today, for the reason mentioned earlier: it'd try to pass an undefined value for the branches rather than leaving it alone.

zaucker:
I add `--> Nil` to the methods of this PR. Will do it for the others once this is merged.

zaucker:
@jnthn For some reason this PR seems to have been stuck on Drone, but now it finished.

jnthn (approved):

PR #169, opened 2020-09-19
Adapt frontend/backend for rename_dataset request

zaucker:


lib/Agrammon/Web/Routes.pm6
```diff
@@ -94,11 +94,13 @@ sub dataset-routes(Agrammon::Web::Service $ws) {
...
-        # test
+        # working
         post -> LoggedIn $user, 'rename_dataset' {
-            request-body -> (:$old!, :$new!) {
-                my $data = $ws.rename-dataset($user, $old, $new);
-                content 'application/json', $data;
+            request-body -> (:oldName($old-name)!, :newName($new-name)!) {
+                my %ret = $ws.rename-dataset($user, $old-name, $new-name);
+                content 'application/json', %(
+                    :oldName(%ret<old>), :newName(%ret<new>)
```
jnthn, line 102: Hm, why is it returning the same thing that was sent in? Can it ever differ?

zaucker: The renaming could fail (because dataset names must be unique). I should probably handle that explicitly.
Would you rather return sucess/failure or so?

jnthn: Yes. I mean, really I'd rather see us use HTTP status codes to indicate failure rather than a 200 OK followed by a body telling us it's not OK. But I'm sympathetic to wanting to at least get things working with the current frontend before such changes.

jnthn: Probably 409 Conflict is the appropriate choice here (and Cro even provides a `conflict` sub for producing that response).

zaucker: I'll give it a try. As I am already making (small) changes to the frontend code, I think it should be ok to handle that properly.

t/webservice.t
```diff
@@ -212,6 +210,42 @@ transactionally {
...
+
+}
+
+transactionally {
+
+    subtest "rename-dataset exception" => {
+        throws-like { $ws.rename-dataset($user, 'TestSingle', 'TestSingle') },
+            X::Agrammon::DB::Dataset::RenameFailed,
+            "Rename dataset fails if old and new name are identical";
+    }
```
jnthn, line 231: A question of taste, but `throws-like` already does a `subtest` call for the things it asserts, so we can have slightly simpler output without the outer layer of `subtest` here.

lib/Agrammon/Web/Routes.pm6
```diff
@@ -60,6 +64,30 @@ sub static-content($root) is export {
...
+                    when X::Agrammon::DB::Dataset::AlreadyExists {
+                        conflict 'application/json', %(
+                            error => .message
+                        );
+                    }
+                    when X::Agrammon::DB::Dataset::RenameFailed {
+                        conflict 'application/json', %(
+                            error => .message
+                        );
+                    }
```
jnthn, line 84: Could be shortened to `when X::Agrammon::DB::Dataset::AlreadyExists | X::Agrammon::DB::Dataset::RenameFailed { ... }` given then `when` bodies are the same.

zaucker: Didn't know how to do that properly, thanks.

t/webservice.t
zaucker: Removed

PR #172, opened 2020-09-19
Adapt frontend/backend for change_password request

zaucker:


lib/Agrammon/DB/User.pm6
```diff
@@ -126,7 +126,7 @@ class Agrammon::DB::User does Agrammon::DB {
                 if not $valid {
                     warn 'PW update failed';
                 }
-                return $valid;
+                return $valid ?? %( :changed(1)) !! False;
```
jnthn, line 129: Hm, bit of an odd return value situation here. Would be better as a straight true/false at the DB level?

lib/Agrammon/Web/Routes.pm6
```diff
@@ -186,10 +186,10 @@ sub dataset-routes(Agrammon::Web::Service $ws) {
 sub user-routes(Agrammon::Web::Service $ws) {
     route {
 
-        # test
+        # working
         post -> LoggedIn $user, 'change_password' {
             request-body -> (:oldPassword($old-password)!, :newPassword($new-password)!) {
-                my $data = $ws.change-password($old-password, $new-password);
+                my $data = $ws.change-password($user, $old-password, $new-password);
                 content 'application/json', $data;
```
jnthn, line 193: When map it into something nice for the response here.

jnthn: Also, should a different HTTP status be used to indicate the error situation, rather than a boolean?

jnthn: But at least it should return the same "shape" of data if a boolean is being used for it.

lib/Agrammon/DB/User.pm6
zaucker: Removed the return value and use 204 instead

lib/Agrammon/Web/Routes.pm6
zaucker: Completely reworked.

lib/Agrammon/Web/Routes.pm6
```diff
@@ -145,6 +145,24 @@ sub api-routes (Str $schema, $ws) {
...
+                    note "$_";
+                    when X::Agrammon::DB::User::PasswordsIdentical {
+                        conflict 'application/json', %(
+                            error => .message
+                        );
+                    }
+                    when X::Agrammon::DB::User::InvalidPassword {
+                        response.status = 422;
+                        response.append-header('Content-type', 'application/json');
+                        response.set-body(to-json { :error(.message) });
```
jnthn, line 161: You can actually just write `content 'application/json', %( error => .message )` (and note you can't use curly syntax here because you reference `$_` which makes it a closure). `content` will only set a `200` status if no status is already set. So in fact `conflict` is implemented by setting status and then calling `content`.

zaucker: Ah, I was hoping for an easier way :-)

jnthn (approved):

PR #174, opened 2020-09-19
Adapt frontend/backend for create_account request

zaucker:


t/webservice.t
```diff
@@ -81,15 +81,12 @@ transactionally {
...
-            :organisation('Org'),
-        );
-        ok $newUser = $ws.create-account($user, $userData), "Create new account";
-        is $newUser.username, 'foo@bar.ch', "User has expected username";
+        ok $username = $ws.create-account(
+            $user,
+            'foo@bar.com', 'myPassword', 'myKey',
+            'Erika', 'Mustermann', 'MyOrg'
+        ), "Create new account";
+        is $username, 'foo@bar.com', "User has expected username";
```
jnthn, line 89: Also add a test for the case where `Any` is passed for the things that are optional, to cover that case.

zaucker: Added another test, some error handling, and OpenAPI validation.

lib/Agrammon/DB/User.pm6
```diff
@@ -15,13 +15,21 @@ class X::Agrammon::DB::User::NoUsername is Exception {
     }
 }
 
+class X::Agrammon::DB::User::CreateFailed is Exception {
+    has Str $.user-name is required;
+
+    method message {
+        "Couldn't create user $.username";
```
jnthn, line 22: Should be `$.user-name` (but you don't really need it to be a virtual call at all, so you can access the underlying attribute directly as `$!user-name`, and those typos are caught at compile time).

lib/Agrammon/Web/Routes.pm6
```diff
@@ -66,6 +66,21 @@ sub static-content($root) {
...
+        # working
+        operation 'createAccount', -> LoggedIn $user {
+            request-body -> (:$email!, :$password!, :$key, :$firstname, :$lastname, :$org) {
+                my $username = $ws.create-account($user, $email, $password, $key, $firstname, $lastname, $org);
+                content 'application/json', { :$username };
+                CATCH {
+                    note "$_";
+                    when X::Agrammon::DB::User::AlreadyExists | X::Agrammon::DB::User::CreateFailed {
+                        conflict 'application/json', %(
+                            error => .message
```
jnthn, line 78: That the bug above isn't caught means we don't have a test covering this line.

share/agrammon.openapi
```diff
@@ -3,6 +3,56 @@ info:
...
+                                - password
+                            properties:
+                                email:
+                                    description: User's email used as username
+                                    type: string
+                                password:
+                                    description: Inital password
+                                    type: string
+                                key:
+                                    description: Validation keu
```
jnthn, line 27: keu?

jnthn (changes requested):

lib/Agrammon/DB/User.pm6
zaucker: Fixed.

lib/Agrammon/Web/Routes.pm6
zaucker: Added test.

share/agrammon.openapi
zaucker: key ... fixed

lib/Agrammon/Web/Routes.pm6
zaucker: With the current setup I don't see that `X::Agrammon::DB::User::CreateFailed` could be thrown without a DB-failure.
I left it in there anyway and added a test for it in `user.t`.

zaucker:
The whole thing got a bit more complex, as I exposed the role attribute as well (not yet used in the frontend, thus optional with default). But now the API is complete

lib/Agrammon/DB/User.pm6
```diff
@@ -30,26 +46,34 @@ class Agrammon::DB::User does Agrammon::DB {
         $!username = $username;
     }
 
-    method create-account(Str $role-name) {
-        warn "*** create account";
-        die X::Agrammon::DB::User::Exists.new(:username($!username)) if self.exists;
+    method create-account($role-name) {
+
+        my $role = $role-name ?? $role-name !! 'user';
```
jnthn, line 51: Or just `$role-name || 'user'`

lib/Agrammon/Web/Routes.pm6
jnthn: Whether all of these should really be 409 is another question, though... I mean, already exists is obviously a conflict. But "unknown role" doesn't really feel like one. It's good we're using a HTTP error code, but perhaps we should try and pick appropriate ones.

share/agrammon.openapi
```diff
@@ -3,6 +3,62 @@ info:
...
+                        application/json:
+                            schema:
+                                type: object
+                                required:
+                                    - username
+                                properties:
+                                    username:
+                                        type: string
+                '409':
+                    description: User already exists
```
jnthn, line 57: And this only documents the conflict case too

lib/Agrammon/DB/User.pm6
zaucker: Done

share/agrammon.openapi
zaucker: I added a few more status codes.

jnthn (approved):

PR #177, opened 2020-09-20
Adapt frontend/backend for delete_instance request

zaucker:


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -329,16 +329,21 @@ class Agrammon::DB::Dataset does Agrammon::DB {
         }
     }
 
-    method delete-input($var-name) {
-        my $instance;
+    method delete-instance($variable-pattern, $instance) {
+        return unless $variable-pattern and $instance;
```
jnthn, line 333: This will return `Nil`, which will be mapped into a JSON `null`. Probably not so desirable?

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -329,16 +329,21 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+                DELETE FROM data_new
+                WHERE data_dataset=dataset_name2id($1,$2) AND data_var LIKE $3
+                                                        AND data_instance = $4
+                RETURNING data_val
+            SQL
 
-        return $instance ?? self!delete-instance-variable($var, $instance)
-                         !! self!delete-variable($var);
+            my $rows = $ret.rows;
+            return $rows if $rows;
```
jnthn, line 345: Also here, probably want to hand back `0` if nothing changed?

lib/Agrammon/Web/Routes.pm6
```diff
@@ -287,10 +287,10 @@ sub application-routes(Agrammon::Web::Service $ws) {
...
 
-        # test/implement
-        post -> LoggedIn $user, 'delete_data' {
-            request-body -> %data {
-                my $ret = $ws.delete-data($user, %data);
+        # working
+        post -> LoggedIn $user, 'delete_instance' {
+            request-body -> ( :datasetName($dataset-name), :$instance, :variablePattern($variable-pattern) )  {
+                my $ret = $ws.delete-instance($user, $dataset-name, $variable-pattern, $instance);
                 content 'application/json', $ret;
```
jnthn, line 294: Returning a plain integer (or null) here, I think. The mock does the wrapping in a `deleted`, but that isn't what the code in `Agrammon::Web::Service` really returns.

lib/Agrammon/DB/Dataset.pm6
zaucker: Added exception handler instead.

lib/Agrammon/DB/Dataset.pm6
zaucker: Removed return value

lib/Agrammon/Web/Routes.pm6
zaucker: Went for 204 and exception handling instead.

jnthn (approved):
In conflict, but approving so it can be resolved and merged after that.

PR #178, opened 2020-09-20
Adapt frontend/backend for rename_instance request

zaucker:


lib/Agrammon/Web/Routes.pm6
```diff
@@ -311,11 +311,11 @@ sub application-routes(Agrammon::Web::Service $ws) {
...
 
-        # test
+        # working
         post -> LoggedIn $user, 'rename_instance' {
-            request-body -> (:datasetName($dataset-name)!, :oldInstance($old-instance)!, :newInstance($new-instance)!, :$pattern!) {
-                my $data = $ws.rename-instance($user, $dataset-name, $old-instance, $new-instance, $pattern);
-                content 'application/json', $data;
+            request-body -> (:datasetName($dataset-name)!, :oldInstance($old-instance)!, :newInstance($new-instance)!, :variablePattern($variable-pattern)!) {
+                my $new-name = $ws.rename-instance($user, $dataset-name, $old-instance, $new-instance, $variable-pattern);
+                content 'application/json', %( :newName($new-name) );
```
jnthn, line 318: What about the error case here? Is `null` OK? (Again, I'd prefer a 409 here.)

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -373,17 +389,29 @@ class Agrammon::DB::Dataset does Agrammon::DB {
                          !! self!delete-variable($var);
     }
 
-    method rename-instance($old-instance, $new-instance, $pattern) {
+    method rename-instance($old-name, $new-name, $pattern) {
```
jnthn, line 392: Minor style point (but it can be done as a PR across the codebase if you agree): when something isn't going to return a useful value, I often add a `--> Nil` at the end of the signature to document/enforce that (enforce as in it means no values leak out and accidentally get used as if they were intended return values).

jnthn (approved):

PR #180, opened 2020-09-20
Adapt frontend/backend for delete_tag request

zaucker:


jnthn:
I guess I have the same question here: how are we handling the thing not existing? Is it approprirate?

lib/Agrammon/Web/Service.pm6
```diff
@@ -82,7 +82,7 @@ class Agrammon::Web::Service {
     }
 
     method delete-tag(Agrammon::Web::SessionUser $user, Str $name) {
-        return Agrammon::DB::Tag.new(:$user, :$name).delete;
+        Agrammon::DB::Tag.new(:$user, :$name).lookup.delete;
```
jnthn, line 85: Hm, I'm a little curious about this construction. So `lookup` looks up the tag by ID. If it's not found, then I guess `$!id` ends up as some undefined value. The we do `delete`. I guess the type object is then mapped into a `NULL` in SQL. But I wonder if this is really the best way to do it; it's two queries rather than one, aside from it working for, if I'm reading correctly, slightly awkward reasons.

jnthn: (I realized this pattern is probably elsewhere too. I'd not thought about it a great deal before (focused more on the route stuff, I guess).

zaucker: I removed the `lookup` here and also handled an exception for deletion failure.
Will look into the other uses of `lookup` separately.

jnthn (approved):

PR #182, opened 2020-09-25
Fix set_tag and remove_tag requests

zaucker:
No return values, exception handling, OpenAPI validation

### DIFF
--- a/lib/Agrammon/DB/Dataset.pm6
+++ b/lib/Agrammon/DB/Dataset.pm6
@@ -165,7 +165,7 @@ class Agrammon::DB::Dataset does Agrammon::DB {
 
 
     #| Set tag on datasets.
-    method set-tag(@datasets!, $tag-name! --> Nil) {
+    method set-tag(@datasets, $tag-name --> Nil) {
         self.with-db: -> $db {
             my $tag-id = Agrammon::DB::Tag.new( :name($tag-name), :$!user).lookup.id;
             die X::Agrammon::DB::Tag::UnknownTag($tag-name) unless $tag-id;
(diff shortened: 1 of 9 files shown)


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -147,31 +163,51 @@ class Agrammon::DB::Dataset does Agrammon::DB {
         return self;
     }
 
-    method set-tag($tagName) {
+
+    #| Set tag on datasets.
+    method set-tag(@datasets!, $tag-name! --> Nil) {
```
jnthn, line 168: Redundant `!` (positional parameter are required by default). Funnily enough, we introduced an annotation to point that out in this week's Comma release. :-)

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -147,31 +163,51 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+                    die X::Agrammon::DB::Dataset::SetTagFailed.new(:$tag-name);
+                }
             }
         }
-        return self;
     }
 
-    method remove-tag($tagName) {
+    #| Remove tag from datasets.
+    method remove-tag(@datasets!, $tag-name! --> Nil) {
```
jnthn, line 190: Same here

jnthn (approved):

lib/Agrammon/DB/Dataset.pm6
zaucker: Removed.

lib/Agrammon/DB/Dataset.pm6
zaucker: Removed

PR #193, opened 2020-10-02
Show outputs with filter groups in frontend results

zaucker:
Also

- use new model in tests
- change config to show Agrammon 6.0

### DIFF
--- a/lib/Agrammon/OutputFormatter/GUI.pm6
+++ b/lib/Agrammon/OutputFormatter/GUI.pm6
@@ -2,9 +2,11 @@ use Agrammon::Model;
 use Agrammon::Outputs;
 
 sub output-for-gui(Agrammon::Model $model,
-                   Agrammon::Outputs $outputs) is export {
+                   Agrammon::Outputs $outputs,
+                   Bool $include-filters
+                   ) is export {
     my %output = %(
-        data => get-data($model, $outputs),
+        data => get-data($model, $outputs, $include-filters),
         log  => %(),
         pid  => 333,
 ### TODO: is this still needed with the new implementation?
@@ -18,25 +20,19 @@ sub output-for-gui(Agrammon::Model $model,
 #    return ();
 #}
 
-sub get-data($model, $outputs) {
+sub get-data($model, $outputs, $include-filters) {
     my @records;
     for sorted-kv($outputs.get-outputs-hash) -> $module, $_ {
         when Hash {
             for sorted-kv($_) -> $output, $raw-value {
-                my $format = $model.output-format($module, $output);
-                my $value = flat-value($raw-value);
-                my $formattedValue = ($format && $value.defined) ?? sprintf($format, $value)
-                                             !! $value;
-                push @records, %(
-                    format    => $format,
-                    print     => $model.output-print($module, $output),
-                    order     => $model.output-order($module, $output),
-                    labels    => $model.output-labels($module, $output),
-                    units     => $model.output-units($module, $output),
-                    fullValue => $value,
-                    value     => $formattedValue,
-                    var       =>  $module ~ '::' ~ $output,
-                );
+                my $var = $module ~ '::' ~ $output;
+                push @records, make-record($module, $output, $model, $raw-value, $var);
+                if $include-filters {
+                    my $value = flat-value($raw-value);
+                    if $raw-value ~~ Agrammon::Outputs::FilterGroupCollection && $raw-value.has-filters {
+                        push-filters(@records, $module, $output, $model, $raw-value, $var);
+                    }
+                }
             }
         }
         when Array {
@@ -44,21 +40,14 @@ sub get-data($model, $outputs) {
                 for sorted-kv(%instance-outputs) -> $fq-name, %values {
                     my $q-name = $module ~ '[' ~ $instance-id ~ ']' ~ $fq-name.substr($module.chars);
                     for sorted-kv(%values) -> $output, $raw-value {
-                        my $format = $model.output-format($fq-name, $output);
-                        my $value = flat-value($raw-value);
-                        my $formattedValue = ($format  && $value.defined) ?? sprintf($format, $value)
-                                                     !! $value;
-                        push @records, %(
-                            format    => $format,
-                            print     => $model.output-print($fq-name, $output),
-                            order     => $model.output-order($fq-name, $output),
-                            labels    => $model.output-labels($fq-name, $output),
-                            units     => $model.output-units($fq-name, $output),
-                            fullValue => $value,
-                            value     => $formattedValue,
-                            var       =>  $q-name ~ '::' ~ $output,
-                        );
-
+                        my $var = $q-name ~ '::' ~ $output;
+                        push @records, make-record($fq-name, $output, $model, $raw-value, $var);
+                        if $include-filters {
+                            my $value = flat-value($raw-value);
+                            if $raw-value ~~ Agrammon::Outputs::FilterGroupCollection && $raw-value.has-filters {
+                                push-filters(@records, $fq-name, $output, $model, $raw-value, $var);
+                            }
+                        }
                     }
                 }
             }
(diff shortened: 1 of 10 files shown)


lib/Agrammon/OutputFormatter/GUI.pm6
```diff
@@ -67,6 +56,48 @@ sub get-data($model, $outputs) {
...
+                 Agrammon::Outputs::FilterGroupCollection $collection,
+                 $var) {
+    my @results = $collection.results-by-filter-group;
+    for @results {
+        my %filters := .key;
+        my $value := .value;
+        my @filters = %filters.map: { .key ~ '=' ~ .value };
+        for @filters.kv -> $idx, $filter-id {
+            push @records, make-record($fq-name, $output, $model, $value, $var, $filter-id);
+            # TODO: what does this do in text formatter? Do we need it here?
```
zaucker, line 92: @jnthn Why is this test for `$idx` needed? Do I need something similar here?

jnthn: @zaucker Not exactly similar because you're producing a data structure rather than text, but you need to do *something*.

A filter may involve multiple key/value pairs (even though at the moment we only ever have one input with `filter = true` in a module). The loop here with the index is making sure that we only output the value and unit once (on the same line as the filter ID), and then if there are other inputs that make up the filter ID those come on the lines below. Basically, it's trying to show composite filter IDs.

You probably don't want the loop here (nor producing the `@filters` array with strings of text), but rather to do something more like:

```
push @records, make-record($fq-name, $output, $model, $value, $var, %filters);
```

Since at the moment the code is producing a string, only to throw a regex at it and parse it again. More deeply, though, it seems that the code in `make-record` only expects there to be one filter, and it only passes along the value. If you want to just do that, then `%filters.values[0]` would give you what you're current sending as `$filter`. Really, depends if you want to do the general thing now (IMHO, I'd just pass the `%filters` hash along to the UI as is and then deal with how to render filters in the frontend).

lib/Agrammon/OutputFormatter/GUI.pm6
```diff
@@ -18,47 +20,34 @@ sub output-for-gui(Agrammon::Model $model,
...
-                            value     => $formattedValue,
-                            var       =>  $q-name ~ '::' ~ $output,
-                        );
-
+                        my $var = $q-name ~ '::' ~ $output;
+                        push @records, make-record($fq-name, $output, $model, $raw-value, $var);
+                        if $include-filters {
+                            my $value = flat-value($raw-value);
+                            if $raw-value ~~ Agrammon::Outputs::FilterGroupCollection && $raw-value.has-filters {
+                                push-filters(@records, $fq-name, $output, $model, $raw-value, $var);
```
jnthn, line 48: I also wonder a bit if the filtered values should be part of the record of the value that they are breaking down, rather than just pushed as extra, flat, records. Maybe it's a bigger refactor than you want to do in the frontend right now, but it would make it easy to do stuff like making the breakdown by filters expandable/collapsable, since you'd know which filter values were the breakdown of which total values quite easily.

lib/Agrammon/OutputFormatter/GUI.pm6
```diff
@@ -67,6 +56,48 @@ sub get-data($model, $outputs) {
...
+        :$var,
+        :$filter,
+    );
+}
+
+sub push-filters(@records, $fq-name, $output, $model,
+                 Agrammon::Outputs::FilterGroupCollection $collection,
+                 $var) {
+    my @results = $collection.results-by-filter-group;
+    for @results {
```
jnthn, line 86: A straight `for $collection.results-by-filter-group {` would avoid a copy of the data into an array and out again (though it's probably not all that performance sensitive here).

lib/Agrammon/OutputFormatter/GUI.pm6
zaucker: I think I'd prefer to do the frontend refactoring later. There is another issue that might be worthwhile doing. Right now I am sending all the outputs to the frontend and the filter them there for the various "reports".
It might be more sensible to only send the outputs on demand, when they are really needed for a report the user actually looks at.
Not really all that relevant as there is not all that much data, so not decided yet what makes most sense.

lib/Agrammon/OutputFormatter/GUI.pm6
zaucker: Ok, thanks. I'll adapt the frontend to handle the filters.

lib/Agrammon/OutputFormatter/GUI.pm6
zaucker: Done, not sure why I did that ...

lib/Agrammon/OutputFormatter/GUI.pm6
zaucker: Fixed

lib/Agrammon/OutputFormatter/GUI.pm6
```diff
@@ -67,6 +56,38 @@ sub get-data($model, $outputs) {
...
+sub push-filters(@records, $fq-name, $output, $model,
+                 Agrammon::Outputs::FilterGroupCollection $collection,
+                 $var) {
+    for $collection.results-by-filter-group {
+        my %filters := .key;
+        my $value := .value;
+        my @filters = %filters.map: { .key ~ '=' ~ .value };
+        for @filters.kv -> $idx, $filter-id {
+            push @records, make-record($fq-name, $output, $model, $value, $var, %filters);
+        }
```
jnthn, line 86: I don't think this makes sense. We're emitting `%filters` into the record (good, as it means we're sending them as structured data). But if it's a complex filter (multiple inputs contribute to the filter) we'll emit the same `%filters` once for each input. We're also doing throwaway work that we don't emit. I think this code should be:

```
    for $collection.results-by-filter-group {
        my %filters := .key;
        my $value := .value;
        push @records, make-record($fq-name, $output, $model, $value, $var, %filters);
    }
```

jnthn (changes requested):

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock/FatteningPigs/Excretion.nhd
```diff
@@ -36,6 +36,32 @@ Flisch R, Sinaj S, Charles R, Richner W 2009. Grundlagen für die die Düngung i
...
+    en = Animal category
+    de = Tierkategorie
+    fr = Catégorie d'animaux
+  ++units
+    en = -
+++description
+    Animal category
+++help
+  +++en
+     <p>There are no sub categories for dairy cows.</p>
```
jnthn, line 59: Ah, and not a code problem, but spotted it: subcategories is one word :)

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock/FatteningPigs/Excretion.nhd
```diff
@@ -36,6 +36,32 @@ Flisch R, Sinaj S, Charles R, Richner W 2009. Grundlagen für die die Düngung i
...
++animalcategory
+  type = enum
+  filter = true
+  default_gui  = fattening_pigs
+  default_calc = fattening_pigs
+  ++enum
+    +++fattening_pigs
+       en = Fattening_pigs
+       de = Mastschweine
+       fr = Porcs à l'engrais
```
jnthn, line 48: Also maybe this doesn't need the underscore in the English output?

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock/FatteningPigs/Excretion.nhd
zaucker: Fixed.

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock/FatteningPigs/Excretion.nhd
zaucker: Removed

lib/Agrammon/OutputFormatter/GUI.pm6
zaucker: Fixed, I am not sure what I was thinking when doing this, if anything ...
Screenshot 2020-10-05 at 21 30 01

jnthn (approved):

PR #197, opened 2020-10-04
Clone dataset

zaucker:
Also refactoring create-dataset to avoid code duplication.

### DIFF
--- a/lib/Agrammon/DB/Dataset.pm6
+++ b/lib/Agrammon/DB/Dataset.pm6
@@ -18,7 +18,7 @@ class X::Agrammon::DB::Dataset::CloneFailed is Exception {
     has Str $.old-username is required;
     has Str $.new-username is required;
     method message {
-        "Dataset '$!old-dataset' of '$!-old-username' couldn't be copied to '$!new-dataset' of '$!new-username'."
+        "Dataset '$!old-dataset' of '$!old-username' couldn't be copied to '$!new-dataset' of '$!new-username'."
     }
 }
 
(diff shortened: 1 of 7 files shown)


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -85,31 +96,62 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
-    has Int  $.records;
+    has Str  $.version is default('2.0-stage'); # TODO: get from config
+    has Int  $.records; # TODO: is this needed?
     has DateTime $.mod-date;
     has $.data;
     has Agrammon::DB::Tag  @.tags;
     has Agrammon::DB::User $.user;
 
-    method create {
+    method create-dataset( $dataset-name, $username, $version, $model ) {
```
jnthn, line 106: If this is factored out for the use of `create` and `clone`, maybe it should be a private method (`method !create-dataset(...)` and the call it as `self!create-dataset(...)`).

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -133,17 +175,11 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
-    method !clone($new) {
-        my $dataset = Agrammon::DB::Dataset.new(:name($new));
-        warn "clone($new) not yet implemented";
-        return $dataset;
     }
 
     method submit($email) {
-        my $new-dataset = self!clone("Clone of $!name");
+        my $new-dataset = 'newDataset';
+#        = self!clone("Clone of $!name");
```
jnthn, line 182: Intentionally left for later?

lib/Agrammon/Web/Routes.pm6
```diff
@@ -105,7 +105,24 @@ sub api-routes (Str $schema, $ws) {
                 }
             }
         }
-        # working
+        operation 'cloneDataset', -> LoggedIn $user {
+            request-body -> ( :oldUsername($old-username), :newUsername($new-username),
```
jnthn, line 109: Should one really be able to specify the username of the user that this is cloned to from the outside? Is creating a dataset on any another user's account a desired feature, or could it be a security issue?

zaucker: This is used for giving a copy of one‘s own dataset(s) to another user.

zaucker: In the other hand, I‘ll have to check that one cannily give one‘s own datasets away.

lib/Agrammon/DB/Dataset.pm6
zaucker: Done.

lib/Agrammon/DB/Dataset.pm6
zaucker: Yes, made it a bit clearer.

jnthn (approved):

PR #199, opened 2020-10-06
Finish detail report on nh3 loss from liquid application

ChHaeni:
- Finished detail report on nh3 loss from liquid application
- Fixed bug in nh3 loss from liquid storage: nh3 loss could be unrealisticly high, higher than the available n given by the n flow into storage. Fix was included on request from T. Kupper
- Included/prepared code for correct calculation of nh3 loss from liquid application. Implementation needs ok from T. Kupper and, possibly, Agrammon-Group 
- Add checks for total n & tan flows into application


ChHaeni:
Bereit für Merge

PR #212, opened 2020-10-08
Add new function to add a scalar to a filter group hash

ChHaeni:


ChHaeni:
Ready to merge

lib/Agrammon/Formula/Builtins.pm6
```diff
@@ -25,6 +25,11 @@ sub get-builtins is export {
                     unless $filter-group ~~ Agrammon::Outputs::FilterGroupCollection;
             $filter-group.scale(+$multiplier)
         },
+        add => -> $filter-group, $additor {
+            die "scale operator expects a filter group as its first argument"
```
jnthn, line 29: To be consistent the error should be "**add** operator expects..."

jnthn (commented):
An automated test would be nice too; I can add one of those if you like, but the place for it is in `output-filter-group-collection.t` and it'd look much like the test in there for `scale`.

jnthn:
(Don't have time today to write the test, but I can get to it tomorrow.)

zaucker:
> (Don't have time today to write the test, but I can get to it tomorrow.)

@jnthn Would it be ok, to merge the PR and you add the test tomorrow? I just don't have time today either.
@ChHaeni Or, of course, if you can add the test, all the better.

jnthn:
@zaucker Sure, don't consider it blocking. Please fix the error text before merge though.

lib/Agrammon/Formula/Builtins.pm6
ChHaeni: fixed in 727e39bccff6c2eb9b9b0c9d132db4cd63ca6943

ChHaeni:
Test added in 57ffd043ffa81da8d19ee6c6d0f846f23ba71d70

t/output-filter-group-collection.t
```diff
@@ -45,6 +45,19 @@ given Agrammon::Outputs::FilterGroupCollection.from-scalar(0) {
...
 
+{
+    my $group = Agrammon::Outputs::FilterGroupCollection.from-filter-to-value-pairs:
+            [{ac => 'blue cow'} => 4, {ac => 'pink cow'} => 7, {ac => 'blue cow'} => 31];
+    given $group.add(2) {
+        isa-ok $_, Agrammon::Outputs::FilterGroupCollection,
+                'Get another filter group back after adding scalar';
+        is +$group, 42, 'Original filter group is not changed in place';
+        is +$_, 48, 'Total numeric value is correct after adding scalar';
+        is-deeply norm(.results-by-filter-group),
```
zaucker, line 56: I think this should be 46, as there are only two different keys in the filter group (`blue cow` and `pink cow`), thus 2 is only added twice.

zaucker (changes requested):
Fix expected value in test for `add()`

t/output-filter-group-collection.t
ChHaeni: I thought that there's no filter grouping when applying such a function. Line 44 says: `'Correct results by filter group after scaling';`, so I guessed there's the scaling and then an extra call to `.results-by-filter-group`? Same for `add`...

ChHaeni: I'm going to change the test to `[{ac => 'blue cow'} => 4, {ac => 'pink cow'} => 7, {ac => 'green cow'} => 31];`, since we never can have a situation where we use add and the filter group contains two identical keys, anyway.

ChHaeni: Fixed in 67d5340fb2593e23b05264e1cfc60459320f99b8

zaucker: The filter groups main purpose is to add up the values with identical keys (for example, animal categories).
The functions dealing with filter groups **must** maintain this property, otherwise their behaviour will be inconsistent.
Therefore, the test for 

> I thought that there's no filter grouping when applying such a function. Line 44 says: `'Correct results by filter group after scaling';`, so I guessed there's the scaling and then an extra call to `.results-by-filter-group`? Same for `add`...

The sentence you quote describes what a successful test will be. The `scale()` function works on an existing filter group, so the order of evaluation is
1. group by filter key
2. multiply the values with the factor for each key.

jnthn: What @zaucker says is accurate. I guess addition of a scalar to a filter group can be useful for something (otherwise, why would you be adding it :-)), but unlike scaling, its semantics are not so neat. With scaling, the grouping of keys does not affect the result, and if you scaled the total you'd get the same outcome. Addition is rather different; I guess if you know your input data and know what result you want, it's fine, but so far as the model of "filter groups are just a value with a breakdown" goes, addition to each key doesn't fit.

PR #214, opened 2020-10-12
Add function to get sign of hash entries

ChHaeni:


t/output-filter-group-collection.t
```diff
@@ -58,6 +58,21 @@ given Agrammon::Outputs::FilterGroupCollection.from-scalar(0) {
...
     }
 }
+
+{
+    my $group = Agrammon::Outputs::FilterGroupCollection.from-filter-to-value-pairs:
+            [{ac => 'blue cow'} => -4, {ac => 'pink cow'} => 0, {ac => 'green cow'} => 31, {ac => 'blue cow'} => 2];
+    given $group.hashsign() {
+        isa-ok $_, Agrammon::Outputs::FilterGroupCollection,
+                'Get another filter group back after adding scalar';
+        is +$group, 29, 'Original filter group is not changed in place';
```
zaucker, line 68: Wrong comment (about scale instead of hashsign), same above and below.

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock.nhd
```diff
@@ -174,19 +206,47 @@ gui      = Livestock,Tierhaltung,Production animale,Livestock
...
-      0 );
-    
+      1 );
+
++test
+  print = check
+  ++formula
+    hashsign( Sum(n_from_poultry_layers_growers_other_poultry,Livestock::Poultry) );
+
+# +is_layers_growers_other_poultry
```
zaucker, line 220: Don't add new code that is commented out. I guess this is not really related to the purpose of this PR.

zaucker:
I don't particularly like the name of this function (we don't talk about hashes anywhere so far, this seems to be an implementation detail). I am not so crazy about `filtergroupSign()` either. Not sure if we'd want a multi-function `sign()`?

May be @jnthn has a better suggestion, I'd like to leave the naming decision to him to keep it consistent.

ChHaeni:
> I don't particularly like the name of this function (we don't talk about hashes anywhere so far, this seems to be an implementation detail). I am not so crazy about `filtergroupSign()` either. Not sure if we'd want a multi-function `sign()`?
> 
> May be @jnthn has a better suggestion, I'd like to leave the naming decision to him to keep it consistent.

Absolutely ok with changing the name of this function, I couldn't think of a nice name, though. Maybe just Sign with capital S?
I have a large number of commits already using this function & I'd rather like to change it once I pushed them if that's ok...

zaucker (commented):
> Absolutely ok with changing the name of this function, I couldn't think of a nice name, though. Maybe just Sign with capital S?
> I have a large number of commits already using this function & I'd rather like to change it once I pushed them if that's ok...

I could live with that.

But could you remove the commented out code of the `.nhd`? It just doesn't seem related to the `hashsign` function. It would be good to have single-purpose PRs (although I must admit that I am sometimes guilty of not following this guideline).

Or, if the code IS related, remove the comments?

> > Not sure if we'd want a multi-function `sign()`?

> Maybe this is my lack of understanding raku. I thought this is necessary to be able to use the function `sign` inside the `hashsign` function...

In raku one can define functions/methods that have more than one implementation and depending on the arguments they are called with, one or another implementation is called. So you could have a `sign(Number)` that would work on a scalar and `sign(Hash)` that would work like your `hashsign` function.

On the other hand, I am not sure if multi-functions/methods are supported in the (limited) subset of raku available for the `.nhd` files. Let's wait for @jnthn on that.

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock.nhd
ChHaeni: Cleaned

zaucker:
Merging, so that @ChHaeni can merge his model changes. @ChHaeni Please change function name if @jnthn has a better suggestion.

jnthn:
> Not sure if we'd want a multi-function sign()?
>
> May be @jnthn has a better suggestion, I'd like to leave the naming decision to him to keep it consistent.

I'd prefer that we just call it `sign`. Raku already offers a `sign` builtin on numbers; we could define a `multi` candidate like:

```
multi sign(Agrammon::Outputs::FilterGroupCollection $group) {
    $group.sign()
}
```

Which implies renaming the method from `hashsign` to `sign` on the `FilterGroupCollection` too. Given that `Builtins.pm6` already has:

```
        sign => &sign,
```

Then so long as we define the `multi` in the `Builtins.pm6` then the `&sign` will include our multi candidate too, and it should all just work out nicely.

PR #221, opened 2020-10-12
Change hashsign -> sign

ChHaeni:


zaucker:
Looks good to me, but I'd like @jnthn to have a look in case I am overlooking something.

jnthn (approved):

PR #223, opened 2020-10-13
Fix grazing calculation

ChHaeni:
- add exp function to builtins
- fix wrong grazing calculation in both models
- add new and remove old technical parameters in cfg files
- correct expected output in test

PR #229, opened 2020-10-14
Fix expected model results

zaucker:
Fix model results from PR #223

### DIFF
--- a/t/run-complex-model.t
+++ b/t/run-complex-model.t
@@ -33,22 +33,30 @@ lives-ok
     },
     'Successfully executed model';
 
-#dd %output<Total><nh3_ntotal>;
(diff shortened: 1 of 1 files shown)


ChHaeni (commented):
Why does this test not fail for model 'hr-inclNOx'? 
'hr-inclNOx' still has the wrong calculation of grazing correction in housing.
What is this test checking?

ChHaeni:
> Why does this test not fail for model 'hr-inclNOx'?
> 'hr-inclNOx' still has the wrong calculation of grazing correction in housing.
> What is this test checking?

Oh, sorry, I see... Only model  'hr-inclNOxExtended' is tested

PR #242, opened 2020-10-15
Add input variable dimensioning barn for fattening pigs

ChHaeni:
- Add input variable dimensioning barn for fattening pigs
- Add correction factor thereof
- Add technical parameters
- Extend data set
- Check in list with newly added input variables


ChHaeni:
Should I capture missing input, so we don't need to extend the database?

zaucker (approved):

zaucker:
> Should I capture missing input, so we don't need to extend the database?

Not sure what you mean.

PR #246, opened 2020-10-16
Create SQL files for variable rename/delete

zaucker:
@ChHaeni I changed the variable listings into SQL command files. Please note the order in the rename (UPDATE)


zaucker:
Merging as there is nothing to test, really.

PR #247, opened 2020-10-16
Start LaTeX creation

zaucker:
First step in creating LaTeX documentation

### DIFF
--- a/lib/Agrammon/Documentation.pm6
+++ b/lib/Agrammon/Documentation.pm6
@@ -1,7 +1,7 @@
 use Agrammon::Model;
 
 #| Create LaTeX document of Model
-sub create-latex( Agrammon::Model $model, :%technical! ) is export {
+sub prepare-model( Agrammon::Model $model, :%technical! ) is export {
 
     my @sections;
     for $model.evaluation-order.reverse -> $module {
@@ -13,7 +13,7 @@ sub create-latex( Agrammon::Model $model, :%technical! ) is export {
 
         my @outputs;
         for $module.output -> $output (:$name, :$formula, :$description, *%) {
-            @outputs.push( %(:name(latex-escape($output.name)), :$formula, :description(latex-escape($output.description)) ) );
+            @outputs.push( %(:name(latex-escape($output.name)), :code($output.code), :$formula, :description(latex-escape($output.description)) ) );
         }
 
         my $tax = latex-escape($module.taxonomy);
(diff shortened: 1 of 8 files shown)


lib/Agrammon/Model.pm6
```diff
@@ -379,4 +380,34 @@ class Agrammon::Model {
         });
         %$!distribution-map-cache
     }
+
+    # Create LaTeX docu
+    method create-docu( :%technical! ) {
```
jnthn, line 385: Is `docu` a LaTeX term, or just being too lazy to type `document`? BTW, if you use `#|` comments instead on routines, then they are doc comments.

jnthn: Also, I'm a bit uncomfortable with this being on `Model`. The model is, well, the core data model of Agrammon; formatting it into a LaTeX description feels like a non-core concept.

lib/Agrammon/Model.pm6
```diff
@@ -379,4 +380,34 @@ class Agrammon::Model {
...
+
+            my @outputs;
+            for $module.output -> $output (:$name, :$formula, :$description, *%) {
+                @outputs.push( %(:name(latex-escape($output.name)), :$formula, :description(latex-escape($output.description)) ) );
+            }
+
+            my $tax = latex-escape($module.taxonomy);
+
+            @sections.push( %(
+                :title("\\subsection\{$tax\}"), :description(latex-escape($module.description)),
```
jnthn, line 402: If you don't want to double backslashes and curlies everywhere but still want to interpolate with scalars, you can build a custom quote construct that only enables scalar variable interpolation: `Q:s"\subsection{$tax}"`

lib/Agrammon/Model.pm6
zaucker: > Is `docu` a LaTeX term, or just being too lazy to type `document`? BTW, if you use `#|` comments instead on routines, then they are doc comments.

Laziness .. renamed ...

zaucker: > Also, I'm a bit uncomfortable with this being on `Model`. The model is, well, the core data model of Agrammon; formatting it into a LaTeX description feels like a non-core concept.

... and moved into its own module.

lib/Agrammon/Model.pm6
zaucker: Thanks, done.

lib/Agrammon/Documentation.pm6
```diff
@@ -0,0 +1,111 @@
+use Agrammon::Model;
+
+#| Create LaTeX document of Model
+sub prepare-model( Agrammon::Model $model, :%technical! ) is export {
+
+    my @sections;
+    for $model.evaluation-order.reverse -> $module {
+
+        my @inputs;
+        for $module.input -> $input (:$name, :$description, *%) {
```
jnthn, line 10: You're not using `$name` and `$description`; destructuring is not particularly cheap, so it's quite some waste if you don't use the result of it (and since we're pulling things out of the model object, I don't see much point in doing it for validation).

lib/Agrammon/Documentation.pm6
```diff
@@ -0,0 +1,111 @@
...
+            @inputs.push( %( :name(latex-escape($input.name)), :description(latex-escape($input.description)) ) );
+        }
+
+        my @outputs;
+        for $module.output -> $output (:$name, :$formula, :$description, *%) {
```
jnthn, line 15: Same comment here.

lib/Agrammon/Documentation.pm6
```diff
@@ -0,0 +1,111 @@
...
+        ));
+
+    }
+    return @sections;
+}
+
+sub create-latex( Str $model-name, Agrammon::Model $model, :%technical! ) is export {
+    my @sections = prepare-model( $model, :%technical );
+
+    say Qs:to/LATEX/;
```
jnthn, line 33: It's probably better design (and a bit faster) to have this build up a string and return it, then `say` the result. (Easy way: make an array of pieces, and then join it at the end.)

jnthn: Can even join them with `\n`.

lib/Agrammon/Documentation.pm6
```diff
@@ -0,0 +1,111 @@
+use Agrammon::Model;
+
+#| Create LaTeX document of Model
+sub prepare-model( Agrammon::Model $model, :%technical! ) is export {
```
jnthn, line 4: If `create-latex` did not produce output you could call it in the test, and then probably not export this implementation detail.

t/documentation.t
```diff
@@ -0,0 +1,30 @@
...
+    'Parsed technical file';
+isa-ok $params, Agrammon::Model::Parameters, 'Correct type for technical data';
+
+lives-ok { prepare-model(
+               $model,
+               technical => %($params.technical.map(-> %module {
+                   %module.keys[0] => %(%module.values[0].map({ .name => .value }))
+               }))
+           );
+}, 'Successfully created docu';
```
jnthn, line 28: I wonder if, for a more robust test, you could feel the LaTeX output to - assuming there is one on path - a latex program, that would at least assert that it is valid LaTeX syntax.

zaucker: I'd like to post-pone this a bit, as I will need to add a `pdflatex` command anyway for the (not yet implemented) PDF report generation. Once that is implemented, that additional test will be trivial.
Added a TODO, though.

lib/Agrammon/Documentation.pm6
zaucker: Indeed. Done ...

lib/Agrammon/Documentation.pm6
zaucker: ... making this trivial. Done.

lib/Agrammon/Documentation.pm6
zaucker: Removed.

lib/Agrammon/Documentation.pm6
zaucker: Removed.

zaucker:
I left the (now not exported) sub `get-model-data` although it could have been easily added to the (now renamed) `create-latex-source` sub to keep getting the data and formatting separated in case the former would have to be modified due to a change in the model. Not sure this really makes sense.

t/documentation.t
jnthn: OK, works for me.

lib/Agrammon/Documentation.pm6
```diff
@@ -1,111 +1,139 @@
...
-    say '\end{document}';
+
+    @latex.push(latex-closing);
+
+    @latex.join("\n");
+}
+
+sub latex-preamble($model-name) {
+    Qs:to/LATEX/;
+        \documentclass[11pt]{article}
```
jnthn, line 99: Ah nice, moving this out to a sub cleans things up a lot.

jnthn (approved):

PR #257, opened 2020-10-20
Fix grazing and yard fluxes for equides

ChHaeni:
- limit n/tan into yard/grazing to total n/tan excreted
- n/tan for shared time yard/grazing is distributed 50%/50%
- fix both models (hr-inclNOxExtended with/without filter group)


zaucker (approved):
@ChHaeni I think there is a place somewhere in the model with a division by (365*24-grazing_days*grazing_hours) which blows up (division by 0).
Not sure if this was fixed (everywhere). Please check at some point.

PR #261, opened 2020-10-20
Indent dump output

zaucker:


lib/Agrammon/Model.pm6
```diff
@@ -306,7 +306,8 @@ class Agrammon::Model {
     method dump {
         my Str $output;
         for @!evaluation-order.reverse {
-            $output ~= $_.taxonomy ~ "\n";
+            my $level = ($_.taxonomy ~~ m:global/ '::' /).elems;
+            $output ~= $_.taxonomy.indent(4 * $level) ~ "\n";
```
jnthn, line 310: `$_.taxonomy` can be spelled `.taxonomy` (in line with the general principle that much of the time, you don't have to explicitly mention `$_`).

lib/Agrammon/Model.pm6
```diff
@@ -306,7 +306,8 @@ class Agrammon::Model {
     method dump {
         my Str $output;
         for @!evaluation-order.reverse {
-            $output ~= $_.taxonomy ~ "\n";
+            my $level = ($_.taxonomy ~~ m:global/ '::' /).elems;
```
jnthn, line 309: This works, but `.taxonomy.comb('::').elems` will too (`comb` being the opposite of `split` - you say what you want).

jnthn (approved):
It'll work, there's an easier way to write it.

lib/Agrammon/Model.pm6
zaucker: I stole that from a few lines above without really thinking if the `$_`  is really needed here.

jnthn (approved):

PR #268, opened 2020-10-21
Add --sort for sorting dump/latex output

zaucker:
I wonder if it should be `@order := ...`, but then the doing the `.reverse` inside the ternary wouldn't work I suppose.

### DIFF
--- a/lib/Agrammon/UI/CommandLine.pm6
+++ b/lib/Agrammon/UI/CommandLine.pm6
@@ -57,16 +57,16 @@ multi sub MAIN('run', ExistingFile $filename, ExistingFile $input, Str $tech-fil
 }
 
 #| Dump model
-multi sub MAIN('dump', ExistingFile $filename, Str :$variants = 'SHL') is export {
-    say chomp dump-model $filename.IO, $variants;
+multi sub MAIN('dump', ExistingFile $filename, Str :$variants = 'SHL', Str :$sort = 'model') is export {
+    say chomp dump-model $filename.IO, $variants, $sort;
 }
 
-multi sub MAIN('latex', ExistingFile $filename, Str $tech-file?, Str :$variants = 'SHL') is export {
-    latex $filename.IO, $tech-file, $variants;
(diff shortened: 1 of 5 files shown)


lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -57,16 +57,16 @@ multi sub MAIN('run', ExistingFile $filename, ExistingFile $input, Str $tech-fil
 }
 
 #| Dump model
-multi sub MAIN('dump', ExistingFile $filename, Str :$variants = 'SHL') is export {
-    say chomp dump-model $filename.IO, $variants;
+multi sub MAIN('dump', ExistingFile $filename, Str :$variants = 'SHL', Str :$sort = 'model') is export {
```
jnthn, line 60: Maybe a `subset` type (or just anonymous `where`) to catch any typos?

lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -57,16 +57,16 @@ multi sub MAIN('run', ExistingFile $filename, ExistingFile $input, Str $tech-fil
...
+    say chomp dump-model $filename.IO, $variants, $sort;
 }
 
-multi sub MAIN('latex', ExistingFile $filename, Str $tech-file?, Str :$variants = 'SHL') is export {
-    latex $filename.IO, $tech-file, $variants;
+multi sub MAIN('latex', ExistingFile $filename, Str $tech-file?, Str :$variants = 'SHL', Str :$sort = 'model') is export {
```
jnthn, line 64: Given it's repeated in a number of places, probably the `subset` type is better.

jnthn (commented):
> I wonder if it should be @order := ..., but then the doing the .reverse inside the ternary wouldn't work I suppose.

Correct, you can't bind a `Seq` to an `@` sigil; you could use an sigilless symbol though (`my \order = ...`), which will retain the laziness of the `reverse` operation.

lib/Agrammon/UI/CommandLine.pm6
zaucker: Done.

zaucker:
Also changed to `\order`, forgot about those.

jnthn (approved):

PR #271, opened 2020-10-21
Add filtering function for filter groups

ChHaeni:


ChHaeni:
Tests to be added...

lib/Agrammon/Formula/Builtins.pm6
```diff
@@ -33,6 +33,22 @@ sub get-builtins is export {
                     unless $filter-group ~~ Agrammon::Outputs::FilterGroupCollection;
             $filter-group.add(+$additor)
         },        
+        # Select only values in filter group 'a' 
+        # where values in filter group 'crit' are greater than 0
+        selectStrict => -> $filter-group-a, $filter-group-crit {
+            die "add operator expects a filter group as its first argument"
+                    unless  $filter-group-a ~~ Agrammon::Outputs::FilterGroupCollection &&
+                            $filter-group-crit ~~ Agrammon::Outputs::FilterGroupCollection;
```
jnthn, line 41: This error text needs updating (and in the other operator), otherwise it can be very confusing when we get errors. Should say "selectStrict function expects its arguments to be filter groups" or similar.

lib/Agrammon/Outputs/FilterGroupCollection.pm6
```diff
@@ -112,6 +112,29 @@ class Agrammon::Outputs::FilterGroupCollection {
...
+        for $other!internal-values-by-filter -> $their-elem {
+            my $their-key = $their-elem.key;
+            if $their-elem.value > $thresh {
+                with %!values-by-filter{$their-key} -> $our-value {
+                    # Push our value if our value > threshold
+                    @result-instances.push: $their-key => $our-value;
+                }
+            } elsif $push-all {
+                # Push threshold if our value <= threshold
+                @result-instances.push: $their-key => $thresh;
```
jnthn, line 131: This branch doesn't do a `with %!values-by-filter{$their-key}`, meaning that if the key does not exist in our filter group, but does exist in the other filter group, then we'll include `$thresh` for it anyway. Is this correct? There's no test that seems to cover this case. Such a test would be like:

```
    my $group-a = Agrammon::Outputs::FilterGroupCollection.from-filter-to-value-pairs:
            [{ac => 'blue cow'} => 4, {ac => 'green cow'} => 31];
    my $group-b = Agrammon::Outputs::FilterGroupCollection.from-filter-to-value-pairs:
            [{ac => 'blue cow'} => 0, {ac => 'green cow'} => 5, {ac => 'pink cow'} => 9];
    given $group-a.select-by-threshold($group-b, 5, True) {
        # ...correct checks here...
    }
```

In this example, the question is if the result should contain a filter key involving pink cow or not? I believe it will, from looking at the code, but I'm not sure - based on the function description - if it should.

jnthn: (And whichever way it is, it's good to test this case.)

lib/Agrammon/Formula/Builtins.pm6
ChHaeni: Damn, I'm so sloppy. Thanks for spotting.

lib/Agrammon/Outputs/FilterGroupCollection.pm6
ChHaeni: Yes it should contain pink cow. Will add the test.

ChHaeni: @jnthn What is the value of `%!values-by-filter{$their-key}` if the key is missing in ours? Is it 0? or not defined?

jnthn: @ChHaeni An undefined value (one that will give `False` in response to `.defined`; `with $value { }` is like `if $value.defined { }`, except it also topicalizes on `$value` (sets the default variable `$_` inside of the block to `$value`).

lib/Agrammon/Formula/Builtins.pm6
ChHaeni: @jnthn sorry, too early for review, need to correct this die message even further

ChHaeni: now, it should be ok

lib/Agrammon/Outputs/FilterGroupCollection.pm6
```diff
@@ -112,6 +112,37 @@ class Agrammon::Outputs::FilterGroupCollection {
...
+                my $our-value = %!values-by-filter{$their-key};
+                if $our-value.defined {
+                    # Push our value if their value > threshold
+                    @result-instances.push: $their-key => $our-value;
+                }
+                else {
+                    # Push 0 if their value > threshold, but key is not existing
+                    # 0 because of missing flow (zero flow)
+                    @result-instances.push: $their-key => 0;                    
+                }
```
jnthn, line 134: This is fine, though can be nice to know for the future that since `with` is a form of `if` you can actually do `with`/`else`.

jnthn (approved):

Issue #270, opened 2020-10-21
Drop entries of hash %b if entries == 0 in hash %a


ChHaeni:
Is it possible to have an operation that drops entries of a filter group hash depending on the entries of another filter group hash?
E.g. we could call this function _filter_:
```
my %capitals = Spain => 'Madrid', 'United States' => 'Washington DC';
my %spanish = Spain => 1, 'United States' => 0;
filter(%capitals, %spanish)
```
would result in a new hash equal to `my %capitals-filtered = Spain => 'Madrid'`

ChHaeni:
Btw, why is the function `addPairwise` defined as:
```
        addPairwise => -> $a, $b {
            my $ag = as-filter-group($a);
            my $bg = as-filter-group($b);
            $ag.apply-pairwise($b, &[+], 0)
        }
```
and not as:
```
        addPairwise => -> $a, $b {
            my $ag = as-filter-group($a);
            my $bg = as-filter-group($b);
            $ag.apply-pairwise($bg, &[+], 0)
        }
```
?

ChHaeni:
Would something like this work?
```
    method filter-pairwise(Agrammon::Outputs::FilterGroupCollection $other, Numeric $thresh) {
        # Process those groups existing on both sides or only the other side.
        my @result-instances;
        my Bool %seen-from-this-instance{FilterKey};
        for $other!internal-values-by-filter -> $their-elem {
            my $their-key = $their-elem.key;
            if ( $their-elem.key > $thresh) {
                with %!values-by-filter{$their-key} -> $our-value {
                    # Exists on both sides.
                    @result-instances.push: $their-key => $our-value.value;
                    %seen-from-this-instance{$their-key} = True;
                }
            }
        }

        Agrammon::Outputs::FilterGroupCollection.new(instances => @result-instances)
    }
```

ChHaeni:
Got it: #271

zaucker:
@ChHaeni Could you describe the use case in Agrammon?

ChHaeni:
Instead of passing an additional variable 'mastercategory' (or similar) we can now simply filter for these super/master categories.
Example:
```
+selectStrict_others
  print = check
  ++units  
    en = kg N/year
    de = kg N/Jahr
    fr = kg N/an
  ++description
    Annual NH3 emission from livestock, master category 'others'.
    Only animal categories, where N input > 0, i.e. categories
    that exist.
  ++formula
    selectStrict(
[...]

PR #273, opened 2020-10-22
Correct upgrade of scalar to filter group in ops

jnthn:
Correct an oversight spotted by @ChHaeni in #270.


zaucker (approved):
Should have caught that reviewing.
Though I had to read Christoph's question three times until I spotted it.

PR #276, opened 2020-10-25
Add filter groups with no inputs to output

zaucker:
- At least some of the reports must show all filter groups (all animal categories), even those that have no inputs.
- Will add a switch for turning this on/off and handling in the frontend in a later PR
Not sure this is the most efficient way to do this.

### DIFF
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -147,6 +147,8 @@ class Agrammon::Model {
     has Agrammon::Model::Module @.evaluation-order;
     has Agrammon::Model::Module @.load-order;
     has ModuleRunner $!entry-point;
+    has %.filter-groups;
+    has %.module-hash;
     has %!output-unit-cache;
     has %!output-print-cache;
     has %!output-format-cache;
@@ -228,6 +230,14 @@ class Agrammon::Model {
 
         my $evaluator = ModuleRunner.new(:$module, :@dependencies);
         %loaded{$module-name} = $evaluator;
+        %!module-hash{$module-name} = $module;
+        for $module.input -> $input {
+            my @keys;
+            next unless $input.is-filter;
+            for $input.as-hash<enum>.keys -> $key {
+                %!filter-groups{ $module-name ~ '::' ~ $input.name  }{$key} = True;
+            }
+        }
         return $evaluator;
     }
 
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -148,7 +148,6 @@ class Agrammon::Model {
     has Agrammon::Model::Module @.load-order;
     has ModuleRunner $!entry-point;
     has %.filter-groups;
-    has %.module-hash;
     has %!output-unit-cache;
     has %!output-print-cache;
     has %!output-format-cache;
@@ -230,7 +229,6 @@ class Agrammon::Model {
 
         my $evaluator = ModuleRunner.new(:$module, :@dependencies);
         %loaded{$module-name} = $evaluator;
-        %!module-hash{$module-name} = $module;
         for $module.input -> $input {
             my @keys;
             next unless $input.is-filter;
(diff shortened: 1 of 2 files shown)


lib/Agrammon/Model.pm6
```diff
@@ -147,6 +147,8 @@ class Agrammon::Model {
     has Agrammon::Model::Module @.evaluation-order;
     has Agrammon::Model::Module @.load-order;
     has ModuleRunner $!entry-point;
+    has %.filter-groups;
+    has %.module-hash;
```
jnthn, line 151: This does not seem to be used anywhere.

lib/Agrammon/Model.pm6
```diff
@@ -228,6 +230,14 @@ class Agrammon::Model {
...
         my $evaluator = ModuleRunner.new(:$module, :@dependencies);
         %loaded{$module-name} = $evaluator;
+        %!module-hash{$module-name} = $module;
+        for $module.input -> $input {
+            my @keys;
+            next unless $input.is-filter;
+            for $input.as-hash<enum>.keys -> $key {
+                %!filter-groups{ $module-name ~ '::' ~ $input.name  }{$key} = True;
+            }
+        }
```
jnthn, line 240: Hmm, so this seems to be building up a global view of all inputs that can serve as filters across all modules, so it's global in nature.

lib/Agrammon/OutputFormatter/JSON.pm6
```diff
@@ -93,10 +93,25 @@ sub make-record($fq-name, $output, $model, $raw-value, $var, :$language, :$print
 sub push-filters(@records, $fq-name, $output, $model,
                  Agrammon::Outputs::FilterGroupCollection $collection,
                  $var) {
+    # model filter groups
+    my %fgs = $model.filter-groups;
```
jnthn, line 97: But the filters are really associated with a particular multi module (perhaps occurring in the modules below it).

lib/Agrammon/OutputFormatter/JSON.pm6
```diff
@@ -93,10 +93,25 @@ sub make-record($fq-name, $output, $model, $raw-value, $var, :$language, :$print
...
+            # remove from model filter group
+            %fgs{$key}{$value}:delete;
+        }
+    }
+
+    # add remaining model filter groups
+    for %fgs.kv -> $fg, %values {
+        for %values.kv -> $key, $value {
+            push @records, make-record($fq-name, $output, $model, Nil, $var, :filters( %( $fg => $key) ) );
+        }
```
jnthn, line 114: So this could end up with entries for filter values that have nothing to do with the module that the filters originated in?

jnthn (commented):
I'm not sure I follow the logic here exactly... Also it may be somewhat duplicating the data held in `Agrammon::Model::FilterSet`.

lib/Agrammon/Model.pm6
zaucker: - removed `module-hash` (currently not used anymore)
- `filter-groups` is used in `Agrammon/OutputFormatter/JSON.pm6`

lib/Agrammon/OutputFormatter/JSON.pm6
zaucker: The filter groups are being used to "attribute" outputs from "anywhere" in the model to the filter groups. E.g., the fluxes and emissions from storage are attributed to the elements of the `animalcategory` filter group.
There is a requirement to show the distribution of those outputs to **all** animal categories (even those not occurring in a given dataset and thus being 0) always in the same order to ease post-processing.

I am not sure how this extends to multiple filter groups. But I guess it would work similarly for example for multiple storage types (if those existed) and one could attribute outputs to those as well or even to both filter groups (by storage type and animal category) and those could be listed similarly in a report showing "empty slots" as well.

lib/Agrammon/Model.pm6
jnthn: Yes, was only talking about `module-hash`.

zaucker:
Closing as @jnthn is doing a new implementation.

PR #277, opened 2020-10-25
Cleanup test DB connection infos

zaucker:


zaucker:
Just some cleanup, merging without review.

PR #290, opened 2020-10-26
Add ?ifnot option to preprocessor

ChHaeni:


zaucker:
If you add few appropriate tests, you can easily find out if this change does the right thing (which on first glance it might).

ChHaeni:
Close PR and re-open in new PR due to GitHub Action issue

PR #291, opened 2020-10-27
Add ?ifnot option to preprocessor

ChHaeni:
PR 'clone' of #290

### DIFF
--- a/lib/Agrammon/Preprocessor.pm6
+++ b/lib/Agrammon/Preprocessor.pm6
@@ -65,12 +65,12 @@ sub preprocess(Str $source, %options --> Str) is export {
                     my $start-line = $number + 1;
                     @open.push(OpenDirective.new(:$start-line, :$enabled, :$matched, :accept-elsif));
                 }
-                when /^ '?elsif' \h+ <option=.ident> \h* $/ {
+                when /^ '?elsif' \h+ [$<negate>='!']? <option=.ident> \h* $/ {
                     if @open.pop -> $prev-part {
                         unless $prev-part.accept-elsif {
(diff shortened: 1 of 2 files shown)


ChHaeni:
Going to add nested tests, too

lib/Agrammon/Preprocessor.pm6
```diff
@@ -59,6 +59,12 @@ sub preprocess(Str $source, %options --> Str) is export {
...
             @result-lines.push('');
             given $content {
+                when /^ '?ifnot' \h+ <option=.ident> \h* $/ {
+                    my $enabled = not so %options{~$<option>};
+                    my $matched = $enabled;
+                    my $start-line = $number + 1;
+                    @open.push(OpenDirective.new(:$start-line, :$enabled, :$matched, :accept-elsif));
+                }
                 when /^ '?if' \h+ <option=.ident> \h* $/ {
                     my $enabled = so %options{~$<option>};
```
jnthn, line 69: I think I'd have preferred this as `?if !A`, because then we can also support `?elsif !A` too without having to introduce an `?elsifnot` which is a bit of a mouthful, plus it's easier to remember two features that can be combined than their product. Also, I think it'd be less code; the existing `when` becomes:

```
when /^ '?if' \h+ [$<negate>='!']? <option=.ident> \h* $/ {
    my $enabled = $<negate> ?? .not !! .so given %options{~$<option>};
```

ChHaeni: I see, much better! I'll change it tomorrow...

jnthn (approved):
Looks good to me. :+1:

PR #292, opened 2020-10-27
Build system

zaucker:
- Initial configure/make setup
- Allows frontend compilation
- Needs more work for packaging for deployment


zaucker:
Merging without review as this is only for local use right now.

PR #293, opened 2020-10-27
Handle model variant config

zaucker:
Separate gui and model variants


jnthn (approved):
Looks OK, don't know if we need anything anywhere to handle missing `variant` setting in the config file? Or is that already taken care of?

zaucker:
> Looks OK, don't know if we need anything anywhere to handle missing `variant` setting in the config file? Or is that already taken care of?

I'll make an issue for config validation.

Issue #299, opened 2020-10-28
Collecting filter groups


zaucker:
- At least some of the reports must show all filter groups (all animal categories), even those that have no inputs.
- The filter groups are being used to "attribute" outputs from "anywhere" in the model to the filter groups. E.g., the fluxes and emissions from storage are attributed to the elements of the animalcategory filter group.
There is a requirement to show the distribution of those outputs to all animal categories (even those not occurring in a given dataset and thus being 0) always in the same order to ease post-processing.

I am not sure how this extends to multiple filter groups. But I guess it would work similarly for example for multiple storage types (if those existed) and one could attribute outputs to those as well or even to both filter groups (by storage type and animal category) and those could be listed similarly in a report showing "empty slots" as well.

See also discussion in chat

zaucker:
Resolved with PR #314

Issue #301, opened 2020-10-28
Cleanup enum inputs


zaucker:
We send `enum`, `options`, and `options-lang` from `Agrammon::Model::Input` to the frontend. One should be enough.

PR #314, opened 2020-10-30
All groups for output

jnthn:
This hopefully provides a solution to #299. If you call `.results-by-filter-group(:all)` (note the new ':all` named argument), then it will include all possible filter keys that could exist (based on the `enum`s in the definitions) rather than just those that have been used in an instance. A `0` value is given to those that have no instances.

This is achieved by dynamic means: filter group collections are constructed with the filter set(s) from the model of all modules that contribute values to the group. This is mostly a case of passing it along, with the exception of pairwise operations which need to perform a union of the module filter sets (thus how a `P+` involving two different modules will end up with the filter sets of both being accounted for).

Hopefully this provides the desired semantics, and also will make it an easy change in the output productions too (I did not update those; @zaucker it's probably best to check out this branch and have a go at updating the output logic and seeing if it does the right thing).

### DIFF
--- a/lib/Agrammon/Outputs/FilterGroupCollection.pm6
+++ b/lib/Agrammon/Outputs/FilterGroupCollection.pm6
@@ -67,7 +67,7 @@ class Agrammon::Outputs::FilterGroupCollection {
                 die "Can only get all values when provenance of the filter group was provided";
             }
             my @results;
-            for $!provenance.keys.sort.reverse -> $filter-set {
+            for $!provenance.keys.sort(*.module.taxonomy).reverse -> $filter-set {
                 for $filter-set.all-possible-filter-keys -> %filters {
                     my $key = FilterKey.new(:%filters);
                     @results.push(%filters => %!values-by-filter{$key} // 0);
--- a/lib/Agrammon/Outputs/FilterGroupCollection.pm6
+++ b/lib/Agrammon/Outputs/FilterGroupCollection.pm6
@@ -64,10 +64,10 @@ class Agrammon::Outputs::FilterGroupCollection {
     method results-by-filter-group(Bool :$all = False) {
         if $all {
             if %!values-by-filter && !$!provenance {
-                die "Can only got all values when provenance of the filter group was provided";
+                die "Can only get all values when provenance of the filter group was provided";
             }
             my @results;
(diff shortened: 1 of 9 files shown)


jnthn:
This also probably helps with #301 in so far as the code in that area is cleaned up by this PR (although it still needs attention).

zaucker:
Will extend the other OutputFormatters in separate PR.

lib/Agrammon/Outputs/FilterGroupCollection.pm6
```diff
@@ -54,30 +57,57 @@ class Agrammon::Outputs::FilterGroupCollection {
...
+    #| group. If :all is passed, then all filters that could possibly have been
+    #| selected will be included, even if no instance used them, and they will
+    #| have a zero value.
+    method results-by-filter-group(Bool :$all = False) {
+        if $all {
+            if %!values-by-filter && !$!provenance {
+                die "Can only get all values when provenance of the filter group was provided";
+            }
+            my @results;
+            for $!provenance.keys.sort.reverse -> $filter-set {
```
jnthn, line 70: Unfortunately, this `sort` is not going to do anything useful. It is a `sort` on instances of `Agrammon::Model::FilterSet`, and will sort on memory location or some such. :-)

jnthn: I did 437fbcc18 but I don't know if it's what you actually want. :-)

lib/Agrammon/Outputs/FilterGroupCollection.pm6
```diff
@@ -54,30 +57,57 @@ class Agrammon::Outputs::FilterGroupCollection {
...
+    #| group. If :all is passed, then all filters that could possibly have been
+    #| selected will be included, even if no instance used them, and they will
+    #| have a zero value.
+    method results-by-filter-group(Bool :$all = False) {
+        if $all {
+            if %!values-by-filter && !$!provenance {
+                die "Can only get all values when provenance of the filter group was provided";
+            }
+            my @results;
+            for $!provenance.keys.sort(*.module.taxonomy).reverse -> $filter-set {
```
zaucker, line 70: This sort does almost what we need. Will make a separate PR for it,

zaucker (approved):

PR #315, opened 2020-10-31
Sort filter groups by load order

zaucker:
While sorting by taxonomy alphabetically gives a reproducable sorting, for the users it will be more logical to have the filter groups sorted by module load order, because this order is also reflected in the GUI.

### DIFF
--- a/lib/Agrammon/ModuleBuilder.pm6
+++ b/lib/Agrammon/ModuleBuilder.pm6
@@ -8,14 +8,14 @@ use Agrammon::Model::Test;
 use Agrammon::Model::Technical;
 
 class Agrammon::ModuleBuilder {
+    has $!module-load-order = 0;
+    
     my constant ORDERED = {
         input => { :enum }
     }
 
-    my Int $module-load-order = 0;
-
     method TOP($/) {
-        make Agrammon::Model::Module.new(|flat($<section>.map(*.ast)).Map, :load-order($module-load-order++));
+        make Agrammon::Model::Module.new(|flat($<section>.map(*.ast)).Map, :load-order($!module-load-order++));
     }
 
     method section:sym<general>($/) {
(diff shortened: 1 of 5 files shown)


lib/Agrammon/ModuleBuilder.pm6
```diff
@@ -12,8 +12,10 @@ class Agrammon::ModuleBuilder {
         input => { :enum }
     }
 
+    my Int $module-load-order = 0;
```
jnthn, line 15: Hm, this is global and not going to be safe if we were to, for example, load to models concurrently (maybe not a big concern right now, but better to do things right than have a subtle bug in the future). Probably should become an instance field and we can `.new` the `ModuleBuilder` actions class in some appropriate place (perhaps holding one per `Agrammon::Model` instance and reusing it).

lib/Agrammon/Outputs/FilterGroupCollection.pm6
```diff
@@ -79,7 +79,18 @@ class Agrammon::Outputs::FilterGroupCollection {
             @results
         }
         else {
-            [%!values-by-filter.map({ .key.filters => .value })]
+#            @results = [%!values-by-filter.map({ .key.filters => .value })]
+            my @results;
+            for $!provenance.keys.sort(*.module.load-order) -> $filter-set {
```
jnthn, line 84: You can't rely on having `$!provenance` set always; if think you'll have to test for it and if it's missing, just do what we originally did, and only if its there then include it.

lib/Agrammon/Outputs/FilterGroupCollection.pm6
```diff
@@ -79,7 +79,18 @@ class Agrammon::Outputs::FilterGroupCollection {
...
+                for $filter-set.all-possible-filter-keys -> %filters {
+                    my $key = FilterKey.new(:%filters);
+                    @results.push(%filters => %!values-by-filter{$key} // 0) if %!values-by-filter{$key}:exists;
```
jnthn, line 87: Or could be `@results.push(%filters => $_) with %!values-by-filter{$key};`

jnthn (approved):

jnthn (approved):

PR #322, opened 2020-11-05
Excel export: add unformatted outputs

zaucker:


lib/Agrammon/OutputFormatter/Excel.pm6
```diff
@@ -5,22 +5,96 @@ use Spreadsheet::XLSX;
...
+#                            }
+                            $row++;
+                        }
+                    }
+                }
+            }
+        }
+        NEXT { # empty line between sections
+            $row++ if $n;
+        }
```
jnthn, line 82: Totally fine, but in case you didn't know: for single-statement phasers you don't need the curlies and can just write `NEXT $row++ if $n`.

lib/Agrammon/OutputFormatter/Excel.pm6
```diff
@@ -5,22 +5,96 @@ use Spreadsheet::XLSX;
...
 multi sub flat-value($value) {
-    $value
+    return $value ~~ Real ?? $value.Real !! $value;
 }
+
 multi sub flat-value(Agrammon::Outputs::FilterGroupCollection $collection) {
-    +$collection
+    my $value = +$collection;
+    $value = $value.Real if $value ~~ Real;
+    return $value;
```
jnthn, line 99: Latest `Spreadsheet::XLSX` fixes what's being worked around here. :)

jnthn (approved):

PR #323, opened 2020-11-05
Require minimum Spreadsheet::XLSX and revert workaround

zaucker:


jnthn (approved):

zaucker:
Merging to see if cache update fixes the problem with the failing Excel test

PR #324, opened 2020-11-09
Add Excel test again

zaucker:
- Enable temporarily disabled test
- Add pdf route test

### DIFF
--- a/t/routes-application.t
+++ b/t/routes-application.t
@@ -123,7 +123,7 @@ subtest 'Get excel export' => {
                 body => {
                     :datasetName('TestSingle'),
                     :language('de'),
-                    :!withFilters
+                    :withFilters
                 }),
                 status => 200;
         };
--- a/t/routes-application.t
+++ b/t/routes-application.t
@@ -1,4 +1,3 @@
-use v6;
 use Agrammon::Web::Routes;
 use Agrammon::Web::Service;
 use Agrammon::Web::SessionUser;
@@ -117,13 +116,12 @@ subtest 'Get output variables' => {
 
 subtest 'Get excel export' => {
     test-service routes($fake-store), :$fake-auth, {
-        test-given '/get_excel_export', {
+        test-given '/export/excel', {
             test post(
                 content-type => 'application/x-www-form-urlencoded',
                 body => {
                     :datasetName('TestSingle'),
                     :language('de'),
-                    :!withFilters
                 }),
                 status => 200;
         };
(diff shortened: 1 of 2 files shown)


zaucker:
It seems the failing `get_excel_export` test in `t/routes-application.t` might be `Cro` problem. It works with version 0.8.3 and fails with 0.8.4.

jnthn:
@zaucker The "Cro problem", assuming it's the same thing I investigated before, turned out to be because the test was wrong but we got away with it because of a now-fixed bug (it didn't validate form-encoded bodies properly before 0.8.4).

t/routes-application.t
```diff
@@ -1,4 +1,3 @@
-use v6;
```
jnthn, line 1: Any reason to remove this, given we recently added it everywhere?

t/routes-application.t
```diff
@@ -42,6 +41,8 @@ my $fake-store = mocked(Agrammon::Web::Service,
         get-output-variables => -> $user, $dataset-name {
             %( :variable('x'), :value(2) )
         },
+        get-pdf-export => -> $user, %params {
+        },
```
jnthn, line 45: If you're not adding any custom result or behavior to the method, and only want to record its being called, you needn't specify an override for it in the mock.

t/routes-application.t
```diff
@@ -115,23 +116,37 @@ subtest 'Get output variables' => {
...
+subtest 'Get excel export' => {
+    test-service routes($fake-store), :$fake-auth, {
+        test-given '/export/excel', {
+            test post(
+                    content-type => 'application/x-www-form-urlencoded',
+                    body => {
+                        :datasetName('TestSingle'),
+                        :language('de'),
+                    }),
+                    status => 200;
```
jnthn, line 128: Maybe worth testing the content type of the response too?

t/routes-application.t
zaucker: Branch from the past ... added again.

t/routes-application.t
zaucker: Ah, didn't know. Removed for the moment.

t/routes-application.t
zaucker: How would I do that? And doesn't OpenAPI enforce that anyway?

jnthn: > How would I do that?

Pass `content-type => '...'` after `status => 200`. [Docs](https://cro.services/docs/reference/cro-http-test#content-type). 

> And doesn't OpenAPI enforce that anyway?

Good point, yes, so it is a bit surplus to requirements. So let's not worry about it.

jnthn (approved):

PR #334, opened 2020-11-25
Change balance tests

ChHaeni:


t/run-complex-model-with-filters.t
```diff
@@ -59,24 +59,91 @@ for <hr-inclNOxExtended hr-inclNOxExtendedWithFilters> -> $model-version {
...
-                'tan_remain_scrubber',  # ntot remaining (vanishing) in air scrubber
-                'n_out_livestock';      # ntot out of housing + yard + grazing
-            for @outputs -> $output {
-                $balance-livestock-ntot .= apply-pairwise(%output-hash<Livestock>{$output}, &infix:<->, 0);
+
+            ### losses for both, TAN and N
+            
+            ## livestock
+            my $losses-livestock-both = %output-hash<Livestock><n_excretion>.scale(0.0);
+                my @outputs-livestock-both =
```
zaucker, line 70: Indentation is wrong. Don't use indentation to visually structure your code, this makes it harder to understand the control flow. Using comments and empty lines to separate "sections" as you already do is fine.

zaucker (changes requested):
There are some rather repetitive parts which should be put into a separate function which is then called with the `@outputs-xxx` array as argument (or even directly with a list of output names, e.g.
```
my $losses-livestock-both = total_losses(
    %output-hash<Livestock>,
    'nh3_nlivestock',       # nh3 loss from housing + yard + grazing
    'n2_ngrazing',          # n2 loss from grazing (housing + yard -> storage)
    'no_ngrazing',          # no loss from grazing (housing + yard -> storage)
    'n2o_ngrazing',         # n2o loss from grazing (housing + yard -> storage)
    'tan_remain_scrubber',  # nh3/tan remaining (vanishing) in air scrubber
);
```
or
```
my $losses-livestock-both = total_losses(
    %output-hash<Livestock>,
   @outputs-livestock-both
);
```
if you prefer (you could define the arrays first and then have a rather compact code section for the calculations).
Both the initialization and for-loop can go into that function.

And I guess there could/should also be a function for the duplicated balance test that is passed the various variables calculated before.

zaucker (approved):

PR #336, opened 2020-11-25
Add new reportDetailed report type

zaucker:
Only show filter group parts in GUI for reports with this type


zaucker:
Frontend change only, merging directly

PR #338, opened 2020-12-02
Detail report

zaucker:
Add detailed reporting showing filter group parts in the GUI
Screenshot 2020-12-02 at 12 41 13


zaucker:
As the backend currently sends all the outputs to the frontend including filter groups, there is a bit of a hack to prevent duplication of lines for the none-detailed report case.
I think I can only get around that by refactoring both frontend and backend to just get the appropriate data for a single report, but that kind if refactoring will have to wait. Left a comment on that.

jnthn (approved):

PR #342, opened 2020-12-03
Sort Excel by labels.sort

zaucker:


lib/Agrammon/OutputFormatter/Excel.pm6
```diff
@@ -70,38 +68,45 @@ sub output-as-excel(
...
                     }
                 }
             }
         }
-        NEXT $row++ if $n; # empty line between sections
     }
 
+    my $col = 0;
+    my $last-module = '';
+    for @records.sort(+*.<order>) -> %rec {
```
jnthn, line 90: I think the `+` here will blow up on your "no order" string:

```
$ raku -e 'say <a b c>.sort(+*)'
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1
```

jnthn: Maybe use 0 or -1 or Inf (depending on if you want the unordered things first or last)

lib/Agrammon/OutputFormatter/Excel.pm6
```diff
@@ -70,38 +68,45 @@ sub output-as-excel(
...
+        if $module ne $last-module {
+            $output-sheet.set($row, $col+0, ~$module, :bold);
+            $last-module = $module;
+            $row++;
+        }
+        $output-sheet.set($row, $col+1, %rec<module>);
+        $output-sheet.set($row, $col+2, %rec<output-label> // 'Output: ???');
+        $output-sheet.set($row, $col+3, %rec<value>, :number-format('#,###'));
+        $output-sheet.set($row, $col+4, %rec<unit-label> // 'Unit: ???');
+        $output-sheet.set($row, $col+5, %rec<order>);
```
jnthn, line 102: Hm, curious that we output the sort order

lib/Agrammon/OutputFormatter/Excel.pm6
```diff
@@ -70,38 +68,45 @@ sub output-as-excel(
...
+    my $col = 0;
+    my $last-module = '';
+    for @records.sort(+*.<order>) -> %rec {
+        %rec<module> ~~ / (.+?) '::' /;
+        my $module = $0 // %rec<module>;
+        if $module ne $last-module {
+            $output-sheet.set($row, $col+0, ~$module, :bold);
+            $last-module = $module;
+            $row++;
+        }
```
jnthn, line 97: Should the sorting be global, or sorted per module? Because it looks like it will be sorting everything without regard to module...

lib/Agrammon/OutputFormatter/Excel.pm6
zaucker: It did blow up and I already fixed it :-) ... I am now using the last order value as default.

lib/Agrammon/OutputFormatter/Excel.pm6
zaucker: Just for debugging (also for Christoph, who needs to tune that.

lib/Agrammon/OutputFormatter/Excel.pm6
zaucker: It probably doesn't matter, but I might change it to by module. Just testing right now.

zaucker: Global sorting works better.

lib/Agrammon/OutputFormatter/Excel.pm6
```diff
@@ -87,6 +90,7 @@ sub input-output-as-excel(
 
     my $col = 0;
     my $last-module = '';
+    my $last-order = -1;
```
jnthn: This particular last-order declaration I see being declared here and assigned to (`$last-order = %rec<order>;`) below, but I don't see it being used anywhere?

zaucker: Already removed in the last commit which probably came after your review.

jnthn (approved):

PR #343, opened 2020-12-03
Sort outputs in GUI and Excel

zaucker:
- fix sorting of outputs to GUI
- handle translations
- sort order of reports parts (store as array)
- simplify and cleanup frontend report generation
- make Excel export generation consistent

### DIFF
--- a/frontend/source/class/agrammon/module/output/Reports.js
+++ b/frontend/source/class/agrammon/module/output/Reports.js
@@ -121,7 +121,7 @@ qx.Class.define('agrammon.module.output.Reports', {
         tcm.setColumnVisible(2,false); // ref value
         tcm.setColumnVisible(4,false); // change
         tcm.setColumnVisible(6,false); // print
-        tcm.setColumnVisible(7,false); // order
+        tcm.setColumnVisible(7,true); // order
         tableModel.setColumnSortable(0,false);
         tableModel.setColumnSortable(1,false);
         tableModel.setColumnSortable(2,false);
@@ -172,7 +172,7 @@ qx.Class.define('agrammon.module.output.Reports', {
             var params = {
                 language     : locale,
                 username     : userName,
-                reports      : this.reportSelected,
+                reports      : this.reportIndex,
                 format       : 'excel',
                 titles       : this.titleSelected,
                 datasetName  : datasetName,
@@ -288,6 +288,7 @@ qx.Class.define('agrammon.module.output.Reports', {
         selectLabel: null,
         selectMenu: null,
         resultData: null,
+        reportIndex: null,
         reportSelected: null,
         titleSelected: null,
         tableModel: null,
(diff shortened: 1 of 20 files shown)


frontend/source/class/agrammon/module/output/Reports.js
```diff
@@ -307,15 +308,15 @@ qx.Class.define('agrammon.module.output.Reports', {
...
-    },
+        __setModelVariant: function(msg) {
+            var model=msg.getData();
+                if (model == 'LU') {
+                this.__submitButton.show();
+                }
+            else {
+                this.__submitButton.exclude();
+                }
+        },
```
jnthn, line 319: Indentation looks a bit dodgy here

lib/Agrammon/LanguageParser.pm6
```diff
@@ -0,0 +1,15 @@
+use v6;
+
+sub parse-lang-values(Str $value --> Hash) is export {
+    my %opt-lang;
+    for (split("\n", $value)) -> $ol {
+        my ($l, $o) = split(/ \s* '=' \s* /, $ol);
+        if not $o {
+            warn "ol=$ol";
```
jnthn, line 8: Maybe make it something like `"Failed to parse language value; ol=$ol"`, so it's easier to understand what the warning is about when it happens in 6 months time. :-)

lib/Agrammon/Model/Result.pm6
```diff
@@ -1,9 +1,44 @@
...
+
 class Agrammon::Model::Result {
     has Str $.name;
     has Str $.type;
-    has  %.selector{Str};
-    has %.data;
+    has %.selector{Str};
+    has @!data-order;
+    has %!data-lookup;
     has Str $._order;
```
jnthn, line 11: I'm a bit curious why the `_` here

lib/Agrammon/UI/Web.pm6
```diff
@@ -36,52 +35,41 @@ class Agrammon::UI::Web {
         my (@reports, @graphs);
 
         for $!model.evaluation-order -> $module {
-            my @module-inputs;
             for $module.results -> $result {
+                my %result-hash = $result.as-hash;
+
                 my $type = $result.type;
                 my %report = %(
```
jnthn, line 42: Missed this before, but if you are assigning pairs into a hash, the `%( ... )` around them is redundant.

frontend/source/class/agrammon/module/output/Reports.js
zaucker: Fixed.

lib/Agrammon/LanguageParser.pm6
zaucker: Indeed ... done.

lib/Agrammon/Model/Result.pm6
zaucker: For some reason I had named it that way in the old version. As there is no occurance of it in the latest `.nhd` files, I changed it to `order`. We might not even need that anymore, but if we do, we can always use the more sensible name.
I also noticed that in two more `.js` files which also will need more refactoring/fixing soon, but for this PR I just changed the `_order`.

lib/Agrammon/UI/Web.pm6
zaucker: Thanks, fixed.

jnthn (approved):

PR #345, opened 2020-12-04
Replace commented out model lines with empty lines in preprocessor

zaucker:
... if there is no reason not to.

### DIFF
--- a/lib/Agrammon/LanguageParser.pm6
+++ b/lib/Agrammon/LanguageParser.pm6
@@ -3,9 +3,10 @@ use v6;
 sub parse-lang-values(Str $value --> Hash) is export {
     my %opt-lang;
     for (split("\n", $value)) -> $ol {
+        next unless $ol;
         my ($l, $o) = split(/ \s* '=' \s* /, $ol);
         if not $o {
-            warn "Failed to parse language value: ol=$ol";
+            warn "Failed to parse language value: ol=\>$ol\<";
             next;
         }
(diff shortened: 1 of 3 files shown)


jnthn:
> ... if there is no reason not to

The reason not to *remove* them is that it makes any line numbers that the parser spits out incorrect. That's why any preproc lines are not just dropped, but rather replaced with an empty line - so when we then parse the result we would give line numbers in any errors that map to the source line numbers. (Yes, it's a hacky way to achieve it, but it's a lot simpler than any of the alternatives!)

zaucker:
> > ... if there is no reason not to
> 
> The reason not to _remove_ them is that it makes any line numbers that the parser spits out incorrect. That's why any preproc lines are not just dropped, but rather replaced with an empty line - so when we then parse the result we would give line numbers in any errors that map to the source line numbers. (Yes, it's a hacky way to achieve it, but it's a lot simpler than any of the alternatives!)

So changing those lines into empty lines would be ok? The reason I did this is that the commented out lines from the language strings in the reports triggered the LanguageParsing problem I added the check and warning for.
And it would also speed up parsing, I suppose.

jnthn:
> So changing those lines into empty lines would be ok? 

Yes.

> The reason I did this is that the commented out lines from the language strings in the reports triggered the LanguageParsing problem I added the check and warning for.

May still need to make it tolerate empty lines.

jnthn (approved):

PR #353, opened 2020-12-09
Add Email module

zaucker:
- Build multi-part message.


lib/Agrammon/Email.pm6
```diff
@@ -0,0 +1,76 @@
...
+    has $!to;
+    has $!from;
+    has $!attachment;
+#    has $!filename;
+    has $!mail;
+
+    submethod TWEAK( :$attachment, :$msg, :$filename, :$to, :$from, :$subject --> Nil) {
+        $!to = $to;
+        $!from = $from;
+        $!subject = $subject;
```
jnthn, line 18: Can do `:$!to, :$!from, :$!subject` in the signature and save these three lines.

lib/Agrammon/Email.pm6
```diff
@@ -0,0 +1,76 @@
...
+                :to([ $!to ]),
+                :message(~$!mail),
+            );
+
+            .quit;
+
+            CATCH {
+                when X::Net::SMTP::Client::Async {
+                    note "Unable to send email message: $_";
+                    .quit
```
jnthn, line 70: Since `CATCH` binds the current exception to `$_`, this will call `.quit` on the wrong thing.

jnthn: Maybe better is above to change it to `LEAVE .quit`, and then you don't need it here at all.

lib/Agrammon/Email.pm6
zaucker: Done.

lib/Agrammon/Email.pm6
zaucker: Done. So this is a bug in the README of https://modules.raku.org/dist/Net::SMTP::Client::Async:cpan:HANENKAMP ?

jnthn: Yes.

lib/Agrammon/Email.pm6
```diff
@@ -0,0 +1,70 @@
...
+
+class Agrammon::Email {
+    has $!msg;
+    has $!subject;
+    has $!to;
+    has $!from;
+    has $!attachment;
+    has $!mail;
+
+    submethod TWEAK( :$attachment, :$msg, :$filename, :$!to, :$!from, :$!subject --> Nil) {
```
jnthn, line 14: Hm, looking at this again, I wonder if some of these should be marked required to ensure correct usage (as nameds are all optional by default, but it looks like only the attachment is really optional).

zaucker: Fixed. Gave the filename a default.

jnthn (approved):

PR #356, opened 2020-12-10
Add inputs and arrange inputs/outputs

zaucker:
- Add, translate and arrange all data in PDF report
- Translate and arrange some data in Excel export
- Add dataset, user, model infos to Excel
Input ordering still missing.


lib/Agrammon/Model.pm6
```diff
@@ -156,30 +158,32 @@ class Agrammon::Model {
...
             }
             my %input-data := $input-data.input-hash-for($!module.taxonomy);
             my %input-defaults := $!module.input-defaults;
             for $!module.input -> Agrammon::Model::Input $input {
-                my $key = $input.name;
+                my $key   = $input.name;
                 my $value = %input-data{$key} // %input-defaults{$key};
-                @result.push: AnnotatedInput.new: :$!module, :$input, :$instance-id, :$value;
+                my $enum  = $input.enum;
+                @result.push: AnnotatedInput.new: :$!module, :$input, :$instance-id, :$value, :$enum, :$gui-root;
```
jnthn, line 186: I'm not sure why we add `enum` as an attribute on the class. It's already accessible via `input`, which is also attached to the class. If you simply want the convenience, just do `has Agrammon::Model::Input $.input is required handles <enum>;` and you'll get a delegating method installed.

lib/Agrammon/OutputFormatter/PDF.pm6
```diff
@@ -133,30 +144,149 @@ sub input-output-as-pdf(
...
 
-    my $pdf = create-pdf(
+    # setup template data
+    %data<outputs>   = @output-formatted;
+    %data<inputs>    = @input-formatted;
+    %data<titles>    = %titles;
+    %data<dataset>   = $dataset-name;
+    %data<username>  = $user.username;
+    %data<model>     = $cfg.gui-variant;
+    %data<version>   = 'Agrammon 6.0';
```
jnthn, line 278: Wonder if this shouldn't come from somewhere more central

lib/Agrammon/OutputFormatter/PDF.pm6
```diff
@@ -133,30 +144,149 @@ sub input-output-as-pdf(
...
+    %data<timestamp> = ~DateTime.now( formatter => sub ($o) {
+        sprintf '%02d.%02d.%04d %02d:%02d:%02d',
+            $o.day, $o.month, $o.year,
+            $o.hour, $o.minute, $o.second,
+    });
```
jnthn, line 283: Really isn't important, but it can be less repetitive to do:

```
@jnthn	Reply…
    %data<timestamp> = ~DateTime.now( formatter => sub ($_) {
        sprintf '%02d.%02d.%04d %02d:%02d:%02d',
            .day, .month, .year,
            .hour, .minute, .second,
    });
```

zaucker: Fixed.

lib/Agrammon/Model.pm6
zaucker: Defnitely not needed, removed.

lib/Agrammon/OutputFormatter/PDF.pm6
zaucker: Getting it from `$cfg` now.

lib/Agrammon/Model.pm6
```diff
@@ -156,30 +157,37 @@ class Agrammon::Model {
         method !annotate-inputs-internal($input-data, %run-already, @result, Str $instance-id?) {
             my $tax = $!module.taxonomy;
             return if %run-already{$tax};
+            my $gui-root-module = $!module.gui-root-module;
+            my %gui-root;
+            if $gui-root-module {
+                # what a silly inline signaling
+                my @g = $gui-root-module.gui.split(',');
+                %gui-root = %( :de(@g[1]), :fr(@g[2]), :en(@g[3] // @g[0])) ;
```
jnthn, line 165: Since you're assigning into a hash, you don't need the `%( ... )` on the right; a list of pairs will do fine.

jnthn (approved):

lib/Agrammon/Model.pm6
zaucker: Ah, I forgot. Just fixed.

PR #359, opened 2020-12-15
Replace Smart table model in Dataset table

zaucker:
- Smart table model is not available in Qooxdoo 6 anymore.
- Makes filtering and sorting of dataset table work again
- Update Qooxdoo compiler
- Add routes for source version of frontend files


zaucker:
Merging as only frontend stuff

PR #361, opened 2020-12-16
More detailed error messages

zaucker:


zaucker:
Frontend only, merging.

PR #366, opened 2020-12-21
Invalidate OutputCache on delete-instance

zaucker:
- `store-data` isn't called when deleting instances, thus be must invalidate the cache explicitly here


zaucker:
@jnthn If the cache invalidation was called in `Route.pm6` instead of in `Service.pm6`, this could be easily tested to be called in `t/routes-application` for example.
I wonder if there is another way to do it with an appropriate use auf `Test::Mock`.

I guess another option would be to have an `invalidate-cache` method in `Agrammon::Web::Service` and call that instead of `$!outputs-cache.invalidate()` and mock `invalidate-cache`. This way this could be tested in both the `delete-instance` and `store-data` route tests.

PR #377, opened 2020-12-28
Fix renamed input parameters

zaucker:
`Poultry::Outdoor::free_range` was renamed to `Poultry::Grazing::free_range`


zaucker:
Wrong dataset changed, closing PR

PR #387, opened 2021-01-07
Use default_calc value as filter value default

### LINKED ISSUE #381: Use default_calc in FilterSets
Inputs without value in the input dataset do not use the specified `default_calc` value:
```
Missing filter value for 'animalcategory' in Livestock::DairyCow::Excretion on instance 'Dairy Cow 1'
  in block  at /media/psf/Home/checkouts/agrammon6/lib/Agrammon/Model/FilterSet.pm6 (Agrammon::Model::FilterSet) line 50
```
The reason is that this is a dummy select box for DairyCows, as there is only a single enum option to "select" from and for convenience there is a `default_calc` and `default_gui` value for it in the model.

jnthn:
Resolves #381.

### DIFF
--- a/lib/Agrammon/Model/FilterSet.pm6
+++ b/lib/Agrammon/Model/FilterSet.pm6
@@ -8,9 +8,12 @@ class Agrammon::Model::FilterSet {
     #| An individual filter field.
     my class Filter {
         has Str $.taxonomy is required;
-        has Str $.input-name is required;
-        has Str $.filter-key = $!taxonomy ~ '::' ~ $!input-name;
+        has Agrammon::Model::Input $.input is required;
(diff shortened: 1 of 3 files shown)


lib/Agrammon/Model/FilterSet.pm6
```diff
@@ -8,9 +8,12 @@ class Agrammon::Model::FilterSet {
     #| An individual filter field.
     my class Filter {
         has Str $.taxonomy is required;
-        has Str $.input-name is required;
-        has Str $.filter-key = $!taxonomy ~ '::' ~ $!input-name;
+        has Agrammon::Model::Input $.input is required;
```
zaucker, line 11: Ah, this is what I missed ...

zaucker (approved):

PR #393, opened 2021-02-07
Fix login and logout

zaucker:
- Add logout route
- Add logout method to `SessionUser`
- Cleanup frontend code a bit
- Misc cleanup

### DIFF
--- a/lib/Agrammon/DB/User.pm6
+++ b/lib/Agrammon/DB/User.pm6
@@ -100,7 +100,7 @@ class Agrammon::DB::User does Agrammon::DB {
         return self;
     }
 
-    method load {
+    method load() {
         die X::Agrammon::DB::User::NoUsername.new unless $!username;
         self.with-db: -> $db {
             my $u = $db.query(q:to/USER/, $!username).hash;
--- a/lib/Agrammon/Web/Routes.pm6
+++ b/lib/Agrammon/Web/Routes.pm6
@@ -454,8 +454,8 @@ sub application-routes(Agrammon::Web::Service $ws) {
         post -> Agrammon::Web::SessionUser $user, 'logout' {
             $user.logout();
             content 'application/json', %(
-                user       => $user.username,
-                sudoUser   => 0
+                :user($user.username),
+                :!sudoUser,
             );
         }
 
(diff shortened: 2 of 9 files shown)


lib/Agrammon/Web/Routes.pm6
```diff
@@ -451,6 +451,14 @@ sub application-routes(Agrammon::Web::Service $ws) {
             content 'application/json', %cfg;
         }
 
+        post -> Agrammon::Web::SessionUser $user, 'logout' {
+            $user.logout();
+            content 'application/json', %(
+                :user($user.username),
+                :!sudoUser,
```
jnthn, line 458: Hm, why is this needed?

jnthn (approved):

lib/Agrammon/DB/User.pm6
```diff
@@ -100,7 +100,7 @@ class Agrammon::DB::User does Agrammon::DB {
         return self;
     }
 
-    method load {
+    method load() {
```
jnthn, line 103: I'm a little curious why this was needed, tbh...they should mean the same thing. What did you observe?

zaucker: > I'm a little curious why this was needed, tbh...they should mean the same thing. What did you observe?

Without `()` after `load` I got in `t/user.t`:

Screenshot 2021-02-08 at 15 19 02

lib/Agrammon/Web/Routes.pm6
zaucker: > Hm, why is this needed?

Do you mean the `:!sudoUser`? If so, that's something that is not implemented yet. The GUI has a `sudo` option for role `admin` so that an admin user can become another user without having to know his password (for supporting "help, my dataset doesn't work"). Upon `logout` the admin user is then returned to his own login account (and thus the GUI needs to know that it has to display the admin's username again and not open the login window.

Not sure if this is a bad idea from a security point of few. But I thought we could discuss this once I have implemented it.

jnthn: Yes, about `sudoUser`. The question was why that could ever matter at logout time; now I understand why!

lib/Agrammon/DB/User.pm6
jnthn: Hm, that sounds like a bug. We'll investigate that.

PR #395, opened 2021-02-07
Gui defaults

zaucker:
- Use current value in various GUI popups, e.g. copy dataset (request Thomas Kupper)
- Use Qooxdoo 6.0.0
- Some cleanups


zaucker:
Frontend only, merging

PR #399, opened 2021-02-11
Branch data

zaucker:
- Some cleanup
- Add storing and loading of branching data (using the current frontend data as is, needs cleanup)
- Add module `t/lib/AgrammonTest`:
  - Extended test database 
  - Sub `transactionally`
  - Removing test databases and sub from other tests in separate PR once this one is merged

### DIFF
--- a/lib/Agrammon/DB/Dataset.pm6
+++ b/lib/Agrammon/DB/Dataset.pm6
@@ -198,7 +198,7 @@ class Agrammon::DB::Dataset does Agrammon::DB {
     method submit($email) {
         my $new-dataset = 'newDataset';
 
-        # TODO: clone dataset; implement sending eMail
+        # TODO: move here from Routes.pm
         warn "Submitting dataset and sending mail for submit($email) not yet implemented";
         return $new-dataset;
     }
@@ -216,7 +216,6 @@ class Agrammon::DB::Dataset does Agrammon::DB {
         }
         return self;
     }
-
 
     #| Set tag on datasets.
     method set-tag(@datasets, $tag-name --> Nil) {
@@ -254,7 +253,7 @@ class Agrammon::DB::Dataset does Agrammon::DB {
                                            WHERE tag_name = $1 )
                        AND tagds_dataset IN ( SELECT dataset_id
                                                 FROM dataset
-                                                WHERE dataset_name = $2 )
+                                               WHERE dataset_name = $2 )
                 SQL
                 CATCH {
                     # other DB failure
@@ -277,40 +276,6 @@ class Agrammon::DB::Dataset does Agrammon::DB {
             $!data = $results.arrays;
         }
         return self;
-    }
-
-    method load-branch-data {
-        warn "*** load-branch-data() not yet implemented";
-        my @data;
-        # self.with-db: -> $db {
-        #     my $username = $!user.username;
-        #     my $results = $db.query(q:to/DATASET/, $username, $!name);
-        #     SELECT data_var, data_val, data_instance_order, branches_data, data_comment
-        #       FROM data_view LEFT JOIN branches ON (branches_var=data_id)
-        #      WHERE data_dataset=dataset_name2id($1,$2)
-        #        AND data_var not like '%::ignore'
-        #      ORDER BY data_instance_order ASC, data_var
-        #     DATASET
-        #     $!data = $results.arrays;
-        # }
-        return @data;
-    }
-
-    method store-branch-data(%data) {
-        warn "*** store-branch-data() not yet implemented";
-        my @data;
-        # self.with-db: -> $db {
-        #     my $username = $!user.username;
-        #     my $results = $db.query(q:to/DATASET/, $username, $!name);
-        #     SELECT data_var, data_val, data_instance_order, branches_data, data_comment
-        #       FROM data_view LEFT JOIN branches ON (branches_var=data_id)
-        #      WHERE data_dataset=dataset_name2id($1,$2)
-        #        AND data_var not like '%::ignore'
-        #      ORDER BY data_instance_order ASC, data_var
-        #     DATASET
-        #     $!data = $results.arrays;
-        # }
-        return @data.keys.elems;
     }
 
     method store-comment($comment) {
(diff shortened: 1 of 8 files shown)


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -542,4 +507,79 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+            SQL
+        }
+
+        for @branch-variables -> %var {
+            my $var-id   = %var<data_id>;
+            my $var-name = %var<data_var>;
+
+            # pg array syntax: { "test_1", "test_2" }
+            my $options-pg = '{' ~ %options{$var-name}.map(*.subst(' ', '_', :g) ).join(',') ~ '}';
+
```
jnthn, line 536: The `DB::Pg` module can handle Postgres arrays;, see the [short Arrays section in the docs](https://modules.raku.org/dist/DB::Pg:cpan:CTILMES). So you should be able to send in a Raku array and have it do the right thing.

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -542,4 +507,79 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+            my $branches = $db.query(q:to/SQL/, |@vars[*;*]).hashes;
+                SELECT branches_data, branches_options
+                  FROM branches
+                 WHERE branches_var in ($1,$2)
+                 ORDER BY branches_id
+            SQL
+            $data = {
+                fractions => $branches[0]<branches_data>,
+                options   => [ $branches[0]<branches_options>, $branches[1]<branches_options> ]
```
jnthn, line 579: Seems here we *do* rely on `DB::Pg` doing the right thing with the arrays.

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -542,4 +507,79 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+            self.with-db: -> $db {
+                my $ret = $db.query(q:to/SQL/, $var-id, $fractions-pg, $options-pg);
+                    INSERT INTO branches (branches_var, branches_data, branches_options)
+                                  VALUES ($1, $2, $3)
+                    ON CONFLICT (branches_var)
+                    DO
+                        UPDATE SET branches_data = EXCLUDED.branches_data,
+                                   branches_options = EXCLUDED.branches_options
+                    SQL
+                die "Couldn't save branch data" unless $ret;
```
jnthn, line 546: Typed exception?

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -542,4 +507,79 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+
+                $db.query(q:to/SQL/, $!user.username, $dataset-name);
+                    UPDATE dataset SET dataset_mod_date = CURRENT_TIMESTAMP
+                     WHERE dataset_id=dataset_name2id($1,$2)
+                SQL
+            }
+
+            return 'Branch data saved';
```
jnthn, line 554: Hm, we don't tend to return messages like this from the DB layer. Storing things seem to mostly give back `Nil` or a row count in this file...

lib/Agrammon/Web/Routes.pm6
```diff
@@ -485,16 +485,16 @@ sub application-routes(Agrammon::Web::Service $ws) {
         # TODO: test/implement load_branch_data()
```
jnthn, line 485: Seems it is implemented now?

lib/Agrammon/Web/Routes.pm6
```diff
@@ -485,16 +485,16 @@ sub application-routes(Agrammon::Web::Service $ws) {
...
         post -> LoggedIn $user, 'load_branch_data' {
             request-body -> (:$name!) {
-                my $data = $ws.load-branch-data($user, $name);
-                content 'application/json', $data;
+                my $ret = $ws.load-branch-data($user, $name);
+                content 'application/json', $ret;
             }
         }
 
         # TODO: test/implement store_branch_data()
```
jnthn, line 493: Also implemented.

t/dataset.t
```diff
@@ -121,74 +124,93 @@ transactionally {
...
+                    ["manure belt without manure belt drying system","0","0","10","7","0","0"],
+                    ["deep pit","13","20","0","0","0","22"],
+                    ["deep litter","0","15","0","8","0","0"]
+                ],
+                "voptions":[
+                    ["manure belt with manure belt drying system","manure belt without manure belt drying system","deep pit","deep litter"],
+                    ["less than twice a month","twice a month","3 to 4 times a month","more than 4 times a month","once a day","no manure belt"]
+                ]
+            }';
+            my $data = from-json $data-json;
```
jnthn, line 175: This works, though I'd probably do it as a Raku array/hash data structure that would come out of `from-json`, so as to have better editor support if I need to tweak the data.

t/dataset.t
```diff
@@ -121,74 +124,93 @@ transactionally {
...
+
+            ok $dataset.store-branch-data(
+                $data<vars>, $data<instance>, $data<options>, $data<data>
+            ), "Store branch data";
```
jnthn, line 179: This would pass with just a true value, which'd be better than the string.

t/lib/AgrammonTest.pm6
```diff
@@ -0,0 +1,105 @@
...
+    SQL
+
+    return 1;
+}
+
+sub transactionally(&test) is export {
+    my $*AGRAMMON-DB-HANDLE = my $db = $*AGRAMMON-DB-CONNECTION.db;
+    $db.begin;
+    test($db);
+    $db.finish;
```
jnthn, line 104: Arguably, `LEAVE $db.finish` would be more robust.

lib/Agrammon/DB/Dataset.pm6
zaucker: Oh, neat. That makes it 0 instead of 11 lines :-)

lib/Agrammon/DB/Dataset.pm6
zaucker: I knew that `SELECT` returns them as arrays. Should have guessed it would work on `INSERT` as well.

lib/Agrammon/DB/Dataset.pm6
zaucker: Removed (it was in the wrong place even, inside the for loop ...).

lib/Agrammon/DB/Dataset.pm6
zaucker: Done.

t/dataset.t
zaucker: I did it like that to be sure I didn't make a mistake converting it from the frontend JSON. But changed it now to use a Raku data structure directly.

t/dataset.t
zaucker: Changed it to `lives-ok`.

t/lib/AgrammonTest.pm6
zaucker: Done.

lib/Agrammon/Web/Routes.pm6
zaucker: I'll make a separate PR for that, cleaning up the data sent from the frontend also.

lib/Agrammon/Web/Routes.pm6
zaucker: Next PR as well.

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -542,4 +515,74 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+
+            my @options = %options{$var-name}.map(*.subst(' ', '_', :g) );
+            self.with-db: -> $db {
+                my $ret = $db.query(q:to/SQL/, $var-id, @fractions[*;*], @options);
+                    INSERT INTO branches (branches_var, branches_data, branches_options)
+                                  VALUES ($1, $2, $3)
+                    ON CONFLICT (branches_var)
+                    DO
+                        UPDATE SET branches_data    = EXCLUDED.branches_data,
+                                   branches_options = EXCLUDED.branches_options
```
jnthn, line 547: Ah, so nice this logic can be done in the database. :-)

lib/Agrammon/Web/Routes.pm6
```diff
@@ -482,19 +483,24 @@ sub application-routes(Agrammon::Web::Service $ws) {
...
-        # TODO: test/implement store_branch_data()
+        # TODO: test
         post -> LoggedIn $user, 'store_branch_data' {
-            request-body -> (:datasetName($dataset-name)!, :%data!) {
-                my $data = $ws.store-branch-data($user, %data, $dataset-name);
-                content 'application/json', $data;
+            request-body -> (:name($dataset-name)!, :%data!) {
+                $ws.store-branch-data($user, $dataset-name, %data);
+                CATCH {
+                    note "$_";
```
jnthn, line 499: Wanted or debugging leftover?

t/dataset.t
```diff
@@ -4,9 +4,13 @@ use Agrammon::DB::Dataset;
 use Agrammon::DB::User;
 use Agrammon::Web::SessionUser;
 use DB::Pg;
+use JSON::Fast;
```
jnthn, line 7: Leftover unused import

jnthn (approved):

t/dataset.t
zaucker: Removed.

lib/Agrammon/Web/Routes.pm6
zaucker: I have those lines in all routes with a `CATCH` so that such failures are recorded in the log file. Is this not a good idea?

lib/Agrammon/DB/Dataset.pm6
zaucker: Yes, I like it. Not quite standardised in all common DBs (PostgreSQL and SQLite both have it, so that's good enough for what we usually use). Found a few articles about potential problems, but at least for Agrammon I don't think they are of concern.

lib/Agrammon/Web/Routes.pm6
jnthn: Ah, right, somehow I thought I was reading this in the DB class; I know that pattern is done here.

Probably Cro can provide something a little nicer for this in the future, but for now it makes sense.

PR #400, opened 2021-02-15
Show short summary results only for admin users

zaucker:
- Hide from normal users until working reliably


zaucker:
Frontend only, merging.

PR #402, opened 2021-02-17
Update to Qooxdoo 6.0.0

zaucker:


zaucker:
Frontend only, merging.

PR #404, opened 2021-02-17
Delete comments

zaucker:
Make it possible to delete (null) comments


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -296,7 +296,7 @@ class Agrammon::DB::Dataset does Agrammon::DB {
              WHERE dataset_id=dataset_name2id($2,$3)
             RETURNING dataset_comment
             SQL
-            $!comment = $comment;
+            $!comment = $comment ?? $comment !! Nil;
```
jnthn, line 299: Can be `$!comment = $comment || Nil`

zaucker: Done

jnthn (approved):

PR #405, opened 2021-02-17
Export error handling

zaucker:
- Show HTML error page

### DIFF
--- a/lib/Agrammon/Web/Routes.pm6
+++ b/lib/Agrammon/Web/Routes.pm6
@@ -511,5 +511,5 @@ sub html-error(%error) {
 }
 
 sub cleanup-filename($filename) {
-    $filename.subst(/<-[\w\s_.-]>/, '', :g);
+    $filename.subst(/<-[\w\ _.-]>/, '', :g);
 }
--- a/lib/Agrammon/Web/Routes.pm6
+++ b/lib/Agrammon/Web/Routes.pm6
@@ -306,7 +306,7 @@ sub api-routes (Str $schema, $ws) {
         operation 'exportExcel', -> LoggedIn $user {
             request-body -> %params {
                 # prevent header injection
-                my $filename = "%params<datasetName>.subst(/<-[\w\s_.-]>/, '', :g).xlsx";
+                my $filename = cleanup-filename "%params<datasetName>.xlsx";
                 my $excel = $ws.get-excel-export($user, %params).to-blob;
                 response.append-header(
                     'Content-disposition',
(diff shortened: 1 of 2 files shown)


lib/Agrammon/Web/Routes.pm6
```diff
@@ -305,36 +305,38 @@ sub api-routes (Str $schema, $ws) {
         }
         operation 'exportExcel', -> LoggedIn $user {
             request-body -> %params {
+                # prevent header injection
+                my $filename = "%params<datasetName>.subst(/<-[\w\s_.-]>/, '', :g).xlsx";
```
jnthn, line 309: Maybe only allow spaces, not all whitespace, otherwise a newline could sneak in (I believe Cro will die if that happens, rather than let the header attack happen, but nice to prevent it here).

lib/Agrammon/Web/Routes.pm6
```diff
@@ -305,36 +305,38 @@ sub api-routes (Str $schema, $ws) {
...
+                my $excel = $ws.get-excel-export($user, %params).to-blob;
                 response.append-header(
```
jnthn, line 311: There's a helper `header` so you can just write this as `header 'Content-disposition', ...;` instead of the `response.append-header` call (they are equivalent, just the first is a little shorter).

lib/Agrammon/Web/Routes.pm6
```diff
@@ -305,36 +305,38 @@ sub api-routes (Str $schema, $ws) {
...
+                        content 'text/html', html-error( %( :error(.message), :$filename ) );
                     }
                 }
             }
         }
         operation 'exportPDF', -> LoggedIn $user {
             request-body -> %params {
+                my $pdf = $ws.get-pdf-export($user, %params);
+                # prevent header injection
+                my $filename = "%params<datasetName>.subst(/<-[\w\s_.-]>/, '', :g).pdf";
```
jnthn, line 329: Same comment, but now I noticed the duplication, maybe the filename calculation wants factoring out.

lib/Agrammon/Web/Routes.pm6
zaucker: Fixed.

lib/Agrammon/Web/Routes.pm6
zaucker: Done.

lib/Agrammon/Web/Routes.pm6
zaucker: Put it into a sub.

jnthn (approved):

PR #409, opened 2021-02-19
Add CLI for create-user and set-password

zaucker:


lib/Agrammon/DB/User.pm6
```diff
@@ -137,15 +137,16 @@ class Agrammon::DB::User does Agrammon::DB {
...
         self.with-db: -> $db {
-            my $uid = $db.query(q:to/USER/, $!username).value;
+            $uid = $db.query(q:to/USER/, $!username).value;
                 SELECT pers_id AS id
                   FROM pers
                  WHERE pers_email = $1
             USER
-            return $uid;
         }
+        return $uid;
```
jnthn, line 149: I'm not sure if you need the variable here at all; doesn't `with-db` evaluate to whatever value as produced by the inner block?

lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -206,20 +231,26 @@ sub run (IO::Path $path, IO::Path $input-path, $technical-file, $variants, $form
     }
 }
 
+sub connect-db(Str $cfg-filename) {
+    my $cfg = Agrammon::Config.new;
+    $cfg.load($cfg-filename);
+    my $db = DB::Pg.new(conninfo => $cfg.db-conninfo);
+    PROCESS::<$AGRAMMON-DB-CONNECTION> = $db;
+    return ($cfg, $db);
```
jnthn, line 239: Given this both loads config and connects the DB, I wonder if a better name is in order...

zaucker: Renamed.

lib/Agrammon/DB/User.pm6
zaucker: Either I am doing something wrong, or this is needed. This code
```
    method exists() {
        my $uid = self.with-db: -> $db {
            $db.query(q:to/USER/, $!username).value;
                SELECT pers_id AS id
                  FROM pers
                 WHERE pers_email = $1
            USER
        }
        note "uid=$uid";
        return $uid;
    }
```
outputs
```console
uid=The 'ForeignCode' class is a Rakudo-specific
implementation detail and has no serviceable parts inside
```

jnthn: Eww, that wants looking at, but not as part of this PR

lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -118,6 +123,26 @@ sub USAGE() is export {
...
+sub create-user($cfg-filename, $username, $firstname, $lastname, $password, $role?) {
+    get-cfg-and-db-handle($cfg-filename);
+    my $user = Agrammon::DB::User.new(
+        :$username, :$firstname, :$lastname, :$password
+    );
+    $user.create-account($role);
+    CATCH {
+        when X::Agrammon::DB::User::Exists  {
+            note "User $username already exists";
+            return;
```
jnthn, line 135: I do wonder if we should exit with a non-zero error code here (in case somebody decides to use this in a script)...

jnthn (approved):
One possible tweak, although not sure how much of a bother it is.

lib/Agrammon/UI/CommandLine.pm6
zaucker: Ok, will add one.

zaucker:
Added none-zero error code. Merging.

PR #411, opened 2021-02-20
Fix parameters

zaucker:
Adapt frontend and route parameters


zaucker:
Mostly frontend; merging.

PR #412, opened 2021-02-21
Fix flattened input order and default value

zaucker:
Cleanup frontend code a bit as well.


zaucker:
Only trivial frontend and backend changes, mostly frontend cleanup. Merging.

PR #413, opened 2021-02-23
Fix branching

zaucker:
- Ignore single-option enums for branching/flattening


zaucker:
Frontend only, merging

PR #418, opened 2021-02-24
Minor GUI Details

zaucker:
- Delayed loading of CHANGELOG
- Hide order columns initially


zaucker:
Frontend only, merging.

PR #419, opened 2021-02-24
Fix instance renaming

zaucker:
- Make sure data changes have the correct changed instance name
- Show sudo for support role
- Show column visibility button for admin role


zaucker:
Frontend only, merging.

PR #426, opened 2021-02-25
Fix spelling of ?if switches

zaucker:


zaucker:
Merging as previously discussed with Christoph

PR #427, opened 2021-02-25
Fix report titles and warning

zaucker:


zaucker:
Warning and model change only, merging

PR #428, opened 2021-02-25
Fix version in CHANGELOG

zaucker:


zaucker:
Just release version and date in CHANGELOG, merging

PR #437, opened 2021-03-06
More input distribution tests

zaucker:
- Test for flattened inputs with space and underscore


zaucker:
@jnthn I think the `sort`ing in the branching test over two different submodules (line 127) is probably not quite right, but I did't see how to do this correctly and efficiently.

zaucker:
> @zaucker I think it's the same problem I had to solve in the input branching tests, and you can see https://github.com/oposs/agrammon/blob/master/t/inputs-branching.t#L158 for how I did it.

Thanks, fixed it.

jnthn (approved):

PR #438, opened 2021-03-14
Copy branching config on dataset clone

zaucker:


zaucker:
PR #443 fixes the problem.

PR #439, opened 2021-03-14
Copy branching data on dataset clone

zaucker:
Forgot to port from Agrammon5

### DIFF
--- a/lib/Agrammon/DB/Dataset.pm6
+++ b/lib/Agrammon/DB/Dataset.pm6
@@ -194,6 +194,7 @@ class Agrammon::DB::Dataset does Agrammon::DB {
 
         my $ds = self!create-dataset( $new-dataset, $new-username, $!version, $!model);
         self.with-db: -> $db {
+            # clone inputs
             $db.query(q:to/SQL/, $ds<id>, $old-username, $old-dataset);
                 INSERT INTO data_new (data_dataset, data_var, data_instance, data_val, data_instance_order, data_comment)
                      SELECT $1, data_var, data_instance, data_val, data_instance_order, data_comment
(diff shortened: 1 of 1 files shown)


lib/Agrammon/DB/Dataset.pm6
```diff
@@ -194,14 +194,43 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+
+            # clone branching data (rows from new and data from old dataset)
+            my $n = 0;
+            for @data -> $data {
+                $db.query(q:to/SQL/, @rows[$n], $data[0], $data[1]);
+                    INSERT INTO branches (branches_var, branches_data, branches_options)
+                       VALUES ($1, $2, $3)
+                SQL
+                $n++;
+            }
```
jnthn, line 230: Could be `for @data.kv -> $n, $data { }` or, perhaps better, `for flat @data Z @rows -> $data, $row { }`.

lib/Agrammon/DB/Dataset.pm6
```diff
@@ -194,14 +194,43 @@ class Agrammon::DB::Dataset does Agrammon::DB {
...
+            # clone inputs
             $db.query(q:to/SQL/, $ds<id>, $old-username, $old-dataset);
                 INSERT INTO data_new (data_dataset, data_var, data_instance, data_val, data_instance_order, data_comment)
                      SELECT $1, data_var, data_instance, data_val, data_instance_order, data_comment
                        FROM data_new
                       WHERE data_dataset = dataset_name2id($2, $3)
             SQL
 
+            # get branched inputs from new dataset
+            my @rows = $db.query(q:to/SQL/, $old-username, $new-dataset).arrays;
```
jnthn, line 206: If it's from the new dataset, should it be `$new-username`?

jnthn (commented):
I do wonder if there's a way to do it all in SQL with an "insert into ... select ..." construct?

zaucker:
> I do wonder if there's a way to do it all in SQL with an "insert into ... select ..." construct?

I am not sure if this would be much easier to understand (speed is not a problem). I think it would have to be a join of two sub-selects with the right order for the `SELECT` part.

The problem is that I need the branching data from the old dataset as source and the IDs from the already copied inputs of the new dataset for the target rows.

May be I should try.

lib/Agrammon/DB/Dataset.pm6
zaucker: That's definitely more elegant, thanks! Done.

jnthn:
> speed is not a problem

In this case we can get away with the approach taken. :-)

lib/Agrammon/DB/Dataset.pm6
zaucker: That's true, I forgot about the case where we copy the dataset into the account of another user.
In this case, however, it can actually be even simpler by using `$ds<id>`. Fixed.

lib/Agrammon/DB/Dataset.pm6
zaucker: I forgot that `kv` on an array gives the index as key.

jnthn: Yes, it's a cute consistency with hashes. :-)

jnthn (approved):

Issue #440, opened 2021-03-15
Distribution auf multiple inputs in branching/flattening


zaucker:
- Inputs to be distributed in branching/flattening are marked in the `.nhd` files with the parameter `branch = true`. This would better be called `distribute = true` or something like that. If renamed, this would have to be done in both the model and the backend. Currently this parameter is always/only set on the input specifying the number of animals (which is not consistently named in the model, but that doesn't really matter).
This is the input going into the variable `$to-dist` in the backend.
- There is another parameter `adjustBranching = true` which is set (consistently?) on the inputs `dimensioning_barn` and not used so far, neither in the model nor in the backend as far as I can tell. The naming of this parameter is also not ideal.
- In a module there should be only one of these two parameters be set (not enforced I think).
- Renaming of these parameters in the model would not be a big effort, as they occur in only 9 modules and at most once each.

The problem is that in distributing those two inputs should **both** be distributed "in parallel". That is, if we have 200 cows and a barn dimension of 200 places and e.g. flatten them into two instances of 50% each, both should be 100 in each instance. Similar for branching.
The agricultural background is that NH3 emissions are depend on the number of animals. However, if there are more places in a barn than animals, those can "wander around" and there excrements are distributed over a larger area as "normal" and as emissions are also a function of surface area, there is an increase of the emissions.
Currently the generated instances would distribute animals only (200 -> 100 + 100), but barn dimension would stay at 200 for both instances and thus a correction would be applied which is actually not real/wanted.

zaucker:
I guess it would be sufficient to have just one parameter instead of two and just distribute all inputs that have this parameter set.

jnthn:
> Inputs to be distributed in branching/flattening are marked in the .nhd files with the parameter branch = true. This would better be called distribute = true or something like that. ...
> This is the input going into the variable $to-dist in the backend.

Given I called it "to distribute" in the backend, I guess I agree. :-)

> There is another parameter adjustBranching = true which is set (consistently?) on the inputs dimensioning_barn and not used so far, neither in the model nor in the backend as far as I can tell. The naming of this parameter is also not ideal.

Yes, seems it's not used at all in the backend.

> In a module there should be only one of these two parameters be set (not enforced I think).

I assume only one of `branching` or `adjustBranching`? When you say "in a module" what do you really mean? I think today it only really makes sense that there's one `branch = true` over all modules that together make up a particular multi-instance module? I ask because...

[...]

PR #442, opened 2021-03-15
Model distribute parameter

zaucker:
First step if issue #440 
- Rename `branch` and `adjustBranching` to `distribute` in model and test models
- Adapt backend

### DIFF
--- a/lib/Agrammon/Model/Input.pm6
+++ b/lib/Agrammon/Model/Input.pm6
@@ -20,19 +20,19 @@ class Agrammon::Model::Input {
     has @!enum-order;
     has %!enum-lookup;
     has Int $.order;
-    has Bool $!branch = False;
+    has Bool $!distribute = False;
     has Bool $!filter = False;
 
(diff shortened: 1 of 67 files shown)


lib/Agrammon/Model/Input.pm6
```diff
@@ -20,19 +20,19 @@ class Agrammon::Model::Input {
...
         }
         with $default_gui {
             $!default-gui = val($_);
         }
-        with $branch {
+        with $distribute {
             if .lc eq 'true' {
-                $!branch = True;
+                $!distribute = True;
             }
```
jnthn, line 36: Coulda just been `$!distribute = .lc eq 'true';`

jnthn (approved):
A tiny cleanup is possible, I'll take it on during my work on the next steps for multiple branches.

lib/Agrammon/Model/Input.pm6
zaucker: Fixed. Will push after rebase.

jnthn: Oops, noticed this after merging. Will do it in my branch.

PR #445, opened 2021-03-16
Minor code optimization

zaucker:


zaucker:
@jnthn is doing it in his next PR.

PR #446, opened 2021-03-16
Allow distribution of multiple inputs when branching

### LINKED ISSUE #440: Distribution auf multiple inputs in branching/flattening

jnthn:
Resolves #440.

### DIFF
--- a/lib/Agrammon/Inputs.pm6
+++ b/lib/Agrammon/Inputs.pm6
@@ -313,19 +313,26 @@ class Agrammon::Inputs::Distribution does Agrammon::Inputs::Storage {
         }
 
         # Get details of the inputs we distribute over, and remove them.
-        die "Distribution over multiple inputs is NYI" if @distribute-over > 1;
-        my $dist-over = @distribute-over[0];
-        my $dist-name = $dist-over.substr($dist-over.rindex('::') + 2);
-        my $dist-taxonomy = $dist-over.substr(0, $dist-over.chars - ($dist-name.chars + 2));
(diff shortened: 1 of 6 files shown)


zaucker (approved):
I'll merge and will look at `t/datasource-db.t` separately.

PR #451, opened 2021-03-30
Fix dataset tagging

zaucker:
- Replace Smart table model
- Cleanup
- DB unique fix


zaucker:
Frontend and minor backend change, merging.

PR #453, opened 2021-03-30
Fix recalc

zaucker:
- ReportSelector wasn't enabled after input changes


zaucker:
Frontend only, merging

PR #454, opened 2021-03-30
Enable result summary

zaucker:
- Might have been fixed with PR #453


zaucker:
Frontend only, merging

PR #456, opened 2021-03-30
Fix root filter complete handling

zaucker:
Get fix from latest Agrammon5 version


zaucker:
Frontend only, merging.

PR #462, opened 2021-04-01
Add mostly french (Google translate) translations

zaucker:


zaucker:
Frontend translations only, merging

PR #470, opened 2021-05-13
Basic integration of Cro::APIToken for providing an API

jnthn:
This is the minimal integration, and can be expanded on from here. It adds:

* A CLI command tool to issue an API token to a user (I didn't fill out the revoke and so forth tools one may want, but added this one as I needed it anyway)
* An implementation of the `Cro::APIToken::Middleware` role that uses the API token metadata to load an appropriate Agrammon user and set it into the request `auth` property
* An example API route

I reworked the routes and middleware application a little to make sure we don't spit out session cookies to API users. Since I didn't have to really run Agrammon for a while with a database, I had to set that up, and also patched a couple of small things that made life a little harder for me than it might have been, and will hopefully ease the way in future setups.

Note that the CI will likely fail as `Cro::APIToken` and `Cro::APIToken::Store::Pg` were only just released to CPAN, and it needs a bit of indexing time.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -11,6 +11,7 @@ CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
 
 COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
 
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
 
 --
 -- Name: dataset_name2id(text, text); Type: FUNCTION; Schema: public; Owner: agrammon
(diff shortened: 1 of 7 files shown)


zaucker (approved):

zaucker:
Shouldn't `Cro::APIToken`, `Cro::APIToken::Manager` and `Cro::APIToken::Store::Pg` be added to the  `depends` section of `META6.json`?

zaucker:
Merging and the trying again with rakudo-21.04

jnthn:
Not quite; it should be `Cro::APIToken` (which is a distribution containing various things) and `Cro::APIToken::Store::Pg`. Fixed now.

PR #473, opened 2021-06-10
Make documentation creation more flexible

zaucker:
Command line option for inclusion/exclusion of docu sections.

### DIFF
--- a/lib/Agrammon/Documentation.pm6
+++ b/lib/Agrammon/Documentation.pm6
@@ -28,7 +28,7 @@ sub create-latex-source( Str $model-name, Agrammon::Model $model, $sort, $doc-se
 
     my $last-section = '';
     for @sections -> %section {
-        next unless @(%section<inputs>).elems or @(%section<outputs>).elems;
+        next unless any %section<inputs outputs>;
 
         my $module  = %section<module>;
         my $section = $module;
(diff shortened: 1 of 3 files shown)


lib/Agrammon/Documentation.pm6
```diff
@@ -27,13 +28,12 @@ sub create-latex-source( Str $model-name, Agrammon::Model $model, $sort, :%techn
 
     my $last-section = '';
     for @sections -> %section {
+        next unless @(%section<inputs>).elems or @(%section<outputs>).elems;
```
jnthn, line 31: Guess I missed this before, but it's probably a little over the top. I assume `%section<inputs>` already is an array, so the `@` is just to strip away the scalar container around it, which one would need to do when iterating it with `for`. But methods always see through the container, so `%section<inputs>.elems` works just fine anyway. But there's more: putting something in boolean context works by calling its `Bool` method, which *also* sees through the container, so `next unless %section<inputs> or %section<outputs>;` will do, at which point we might shorten it to `next unless any %sections<inputs outputs>`.

lib/Agrammon/Documentation.pm6
```diff
@@ -18,7 +19,7 @@ sub get-model-data( Agrammon::Model $model, $sort, :%technical! ) {
 }
 
 #| Create a LaTeX source of the model
-sub create-latex-source( Str $model-name, Agrammon::Model $model, $sort, :%technical! ) is export {
+sub create-latex-source( Str $model-name, Agrammon::Model $model, $sort, $doc-sections, :%technical! ) is export {
```
jnthn, line 22: The call to this in `documentation.t` needs updating

zaucker: Oh, silly, ...

jnthn (approved):

PR #476, opened 2021-06-11
Implement input validation against the model

jnthn:
This catches a wide range of problems, including:

* Inputs for modules that don't exist
* Single-instance inputs for multi-instance modules and vice versa
* Missing inputs that don't have a default_calc
* Inputs whose value does not match the type
* Inputs for enum types where the value does not fall within the enum
* Inputs with range constraints (`between`, `ge`) that are not met by
  the value

For multi-instance modules, the instance at fault is also reported in
the error. A list of error objects is produced. This is not yet in any
way integrated with the overall workflow. Since we weren't validating
these things before, it's possible that starting to do so will turn up
genuine mistakes in input data. It's also possible more realistic data
will turn up problems in this validator, although hopefully the decent
level of test coverage here will mean we don't see too much of that.

### DIFF
--- a/lib/Agrammon/Validation.pm6
+++ b/lib/Agrammon/Validation.pm6
@@ -1,0 +1,247 @@
+unit module Agrammon::Validation;
+use Agrammon::Inputs;
+use Agrammon::Model;
+
+#| The root of all validation problems found.
+role Problem {
+    #| The module where the problem was found.
+    has Str $.module is required;
+}
+
+#| An input value for a module that doesn't exist at all.
+class NoSuchModule does Problem {
+    #| The instance, if any.
+    has Str $.instance;
+
+    method message() {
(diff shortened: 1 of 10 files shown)


lib/Agrammon/Validation.pm6
```diff
@@ -0,0 +1,247 @@
...
+
+#| An input value for a module that doesn't exist at all.
+class NoSuchModule does Problem {
+    #| The instance, if any.
+    has Str $.instance;
+
+    method message() {
+        "Input for module '$!module'" ~
+                ($!instance ?? " of instance $!instance" !! "") ~
+                ", but not such module exists in the model"
```
zaucker: typo: not => no

zaucker (approved):
I fixed the typo directly by editing the file in GitHub. I hope this works as expected :-)

PR #485, opened 2021-06-15
Allow reading CSV from STDIN

zaucker:


lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -394,3 +390,9 @@ sub web(Str $cfg-filename, Str $model-filename, Str $technical-file?) is export
 sub preprocessor-options(Str $variants) {
     set($variants.split(","));
 }
+
+sub get-input-filehandle(IO::Path $path) {
+    $path eq '-' ?? $*IN
+                 !! open $path, :r
+                        or die "Couldn't open file $path for reading";
```
jnthn, line 397: I'm not sure if you need this code:

```
$ echo "hi" | raku -e 'my $fh = open "-", :r; say $fh.slurp'
hi
```

jnthn: (That is, it handles the `-` meaning STDIN by default)

zaucker: Not right now, but https://docs.raku.org/type/IO::Handle says:
```
As of 6.d language version, the use of path '-' is deprecated and it will be removed in future language versions entirely.
```

jnthn: Hmm, curious. Maybe somebody thought it was too magical...

jnthn (approved):

PR #486, opened 2021-06-17
Improve and fix output formatters

zaucker:
- Add filtering where missing
- Add filter group output to CSV
- Add tests for `--include-filter` outputs

### DIFF
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -455,7 +455,7 @@ class Agrammon::Model {
 
     method should-print(Str $module, Str $output, @print-set) {
         %!output-print-cache ||= @!evaluation-order.map({
-            .taxonomy => %(.output.map({ .name => (.print//'').split(',').List }))
+            .taxonomy => %(.output.map({ .name => (.split(',').List with .print) }))
         });
         !@print-set or ( %!output-print-cache{$module}{$output} // () ) ∩ @print-set
     }
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -457,7 +457,7 @@ class Agrammon::Model {
         %!output-print-cache ||= @!evaluation-order.map({
             .taxonomy => %(.output.map({ .name => (.split(',').List with .print) }))
         });
-        !@print-set or ( %!output-print-cache{$module}{$output} // () ) ∩ @print-set
+        !@print-set or ( %!output-print-cache{$module}{$output} ) ∩ @print-set
     }
 
     method output-format(Str $module, Str $output) {
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -216,6 +216,7 @@ class Agrammon::Model {
     has ModuleRunner $!entry-point;
     has %!output-unit-cache;
     has %!output-print-cache;
+    has %!output-print-str-cache;
     has %!output-format-cache;
     has %!output-labels-cache;
     has %!output-order-cache;
@@ -446,10 +447,17 @@ class Agrammon::Model {
     }
 
     method output-print(Str $module, Str $output) {
-        %!output-print-cache ||= @!evaluation-order.map({
+        %!output-print-str-cache ||= @!evaluation-order.map({
             .taxonomy => %(.output.map({ .name => .print }))
         });
-        return %!output-print-cache{$module}{$output} // ''
+        return %!output-print-str-cache{$module}{$output} // ''
+    }
+
+    method should-print(Str $module, Str $output, @print-set) {
+        %!output-print-cache ||= @!evaluation-order.map({
+            .taxonomy => %(.output.map({ .name => (.print//'').split(',').List }))
+        });
+        !@print-set or ( %!output-print-cache{$module}{$output} // () ) ∩ @print-set
     }
 
     method output-format(Str $module, Str $output) {
--- a/lib/Agrammon/OutputFormatter/CSV.pm6
+++ b/lib/Agrammon/OutputFormatter/CSV.pm6
@@ -1,46 +1,97 @@
 use v6;
 use Agrammon::Model;
 use Agrammon::Outputs;
+use Agrammon::Outputs::FilterGroupCollection;
 
 sub output-as-csv(
-    $simulation-name, $dataset-id,
+    $simulation-name,
+    $dataset-id,
     Agrammon::Model $model,
     Agrammon::Outputs $outputs,
-    Str $unit-language,
-    Bool $include-filters?, Bool :$all-filters = False --> Str
+    Str $language,
+    @print-set,
+    Bool $include-filters,
+    Bool :$all-filters = False --> Str
 ) is export {
-    # TODO: handle with-filters
-    return (gather for $outputs.get-outputs-hash.sort(*.key) {
-        my $module = .key;
+    return (gather for sorted-kv($outputs.get-outputs-hash) -> $module, $_ {
         my $prefix = "$simulation-name;$dataset-id" if $simulation-name;
-        if .value.isa(Hash) {
-            for .value.sort(*.key) {
+        when Hash {
+            for sorted-kv($_) -> $output, $value {
+                my $var-print = $model.output-print($module, $output) ~ ',All';
+                next unless $var-print.split(',') ∩ @print-set;
+
+                my $filters = '';
+                my $raw-value = $value;
+                my $var = $output;
+                my $unit = $model.output-unit($module, $output, $language);
                 my @data = (
-                    $module, .key, flat-value(.value) // '',
-                    $model.output-unit($module, .key, $unit-language)
+                    $module,
+                    $var,
+                    $filters,
+                    flat-value($raw-value) // '',
+                    $unit
                 );
                 @data.unshift($prefix) if $prefix;
                 take @data.join(';');
-            }
-         }
-        else {
-            for .value.sort(*.key) {
-                my $instance-id = .key;
-                for .value.sort(*.key) {
-                    my $fq-name = .key;
-                    my $q-name = $module ~ '[' ~ $instance-id ~ ']' ~ $fq-name.substr($module.chars);
-                    for .value.sort(*.key) {
-                        my @data = (
-                            $q-name, .key, flat-value(.value) // '',
-                            $model.output-unit($fq-name, .key, $unit-language)
-                        );
-                        @data.unshift($prefix) if $prefix;
-                        take @data.join(';');
+                if $include-filters {
+                    if $raw-value ~~ Agrammon::Outputs::FilterGroupCollection && $raw-value.has-filters {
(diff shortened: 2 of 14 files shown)


lib/Agrammon/OutputFormatter/CSV.pm6
```diff
@@ -1,46 +1,97 @@
...
+        my @filters = %filters.map: { .key ~ '=' ~ .value };
+        for (@filters || '(Uncategorized)').kv -> $idx, $filter-id {
+            my @data = (
+                $q-name,
+                $fq-name,
+                $filters,
+                flat-value($raw-value) // '',
+                $unit
+            );
+            @data.unshift($prefix) if $prefix;
```
jnthn, line 91: Alternatively you can slip it into the original array construction if `|($prefix if $prefix)`

lib/Agrammon/OutputFormatter/CSV.pm6
```diff
@@ -1,46 +1,97 @@
...
+
+sub add-filters(Agrammon::Outputs::FilterGroupCollection $collection,
+        $q-name, $fq-name, $unit, Str $prefix, Bool :$all-filters) {
+    my @results = $collection.results-by-filter-group(:all($all-filters));
+    for @results {
+        my %filters := .key;
+        my $raw-value := .value;
+        my $filters = %filters.values.join(',');
+        my @filters = %filters.map: { .key ~ '=' ~ .value };
+        for (@filters || '(Uncategorized)').kv -> $idx, $filter-id {
```
jnthn, line 83: Both `$idx` and `$filter-id` are unused in the loop?

lib/Agrammon/OutputFormatter/JSON.pm6
```diff
@@ -6,36 +6,38 @@ use Agrammon::Outputs::FilterGroupCollection;
...
-    return get-data($model, $outputs, $include-filters, $language, $prints);
+    return get-data($model, $outputs, $include-filters, @print-set, $language);
 }
 
 sub output-for-gui(Agrammon::Model $model,
                    Agrammon::Outputs $outputs,
                    :$include-filters,
                    :$language
                    ) is export {
+    my @print-set = ('All');
```
jnthn, line 22: Unimportant, but don't need the parens here in Raku (and in a few other places in this PR).

lib/Agrammon/OutputFormatter/Text.pm6
```diff
@@ -5,28 +5,34 @@ use Agrammon::Outputs;
...
         when Hash {
             for sorted-kv($_) -> $output, $value {
-                my $val = flat-value($value // 'UNDEFINED');
                 my $var-print = $model.output-print($module, $output) ~ ',All';
-                if not defined $reports-selected or $var-print.split(',') ∩ @print-set {
-                    $n++;
-                    my $unit = $model.output-unit($module, $output, $language);
+                next unless $var-print.split(',') ∩ @print-set;
+
+                $n++;
```
jnthn, line 24: I'm trying to figure out what `$n` is for here. It seems to be it will always be non-zero if `@module-lines` has any data, and later we do `@lines.append: @module-lines if $n;` but if `@module-lines` is empty this is a no-op.

lib/Agrammon/OutputFormatter/Text.pm6
```diff
@@ -84,6 +88,13 @@ sub render-filters(@module-lines, Agrammon::Outputs::FilterGroupCollection $coll
     }
 }
 
+multi sub flat-value($value) {
+    $value
+}
+multi sub flat-value(Agrammon::Outputs::FilterGroupCollection $collection) {
+    +$collection
+}
+
```
jnthn, line 97: I do wonder if we should put this `and sorted-kv` into some kind of `Agrammon::OutputUtil` module so we can re-use them...

zaucker: Done.

lib/Agrammon/OutputFormatter/JSON.pm6
zaucker: Removed

lib/Agrammon/OutputFormatter/CSV.pm6
```diff
@@ -1,46 +1,97 @@
...
+        }
+        when Array {
+            for sorted-kv($_) -> $instance-id, %instance-outputs {
+                my $filters = '';
+                for sorted-kv(%instance-outputs) -> $fq-name, %values {
                     my $q-name = $module ~ '[' ~ $instance-id ~ ']' ~ $fq-name.substr($module.chars);
-                    for .value.sort(*.key) {
+                    for sorted-kv(%values) -> $output, $value {
+                        my $var-print = $model.output-print($fq-name, $output) ~ ',All';
+                        next unless $var-print.split(',') ∩ @print-set;
```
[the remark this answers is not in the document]
zaucker: Refactored. Turns out, I need it both as String and List

lib/Agrammon/OutputFormatter/CSV.pm6
zaucker: Removed

lib/Agrammon/OutputFormatter/CSV.pm6
zaucker: Done

lib/Agrammon/OutputFormatter/Text.pm6
zaucker: The idea is, that if we have a module, that has no outputs, then the title line (with the module name) should **not**. be added to the output. `@module-lines` is never empty (it always has at least the title or in case of multi-modules also a sutitle), but if there is no output, `$n == 0`.

I refactored it to use a separate array for the titles and now can check `if @module-lines` without the need for `$n`.

I also renamed `@module-lines` to `@output-lines`.

lib/Agrammon/Model.pm6
```diff
@@ -446,10 +447,17 @@ class Agrammon::Model {
...
         });
-        return %!output-print-cache{$module}{$output} // ''
+        return %!output-print-str-cache{$module}{$output} // ''
+    }
+
+    method should-print(Str $module, Str $output, @print-set) {
+        %!output-print-cache ||= @!evaluation-order.map({
+            .taxonomy => %(.output.map({ .name => (.split(',').List with .print) }))
+        });
+        !@print-set or ( %!output-print-cache{$module}{$output} // () ) ∩ @print-set
```
jnthn, line 460: I'm wondering if we ever have a case where we won't have an entry in the hash now? Unlike with the string case, we'll store `Empty` under outputs that have no prints, so the `// ()` is probably redundant.

lib/Agrammon/OutputFormatter/CSV.pm6
```diff
@@ -1,51 +1,94 @@
...
+                            if $raw-value ~~ Agrammon::Outputs::FilterGroupCollection && $raw-value.has-filters {
+                                add-filters($raw-value, $q-name, $fq-name, $unit, $prefix, :$all-filters);
+                            }
+                        }
                     }
                 }
             }
         }
-    }).join("\n") ~ "\n";
+    }).join("\n");
```
jnthn, line 71: Is everything upstream from this OK about losing the trailing `\n`?

jnthn: (For example, we don't get a regression anywhere in output writing of multiple datasets?)

lib/Agrammon/OutputFormatter/CSV.pm6
```diff
@@ -1,51 +1,94 @@
...
+            my @data = (
+                |($prefix if $prefix),
+                $q-name,
+                $fq-name,
+                $filters,
+                flat-value($raw-value) // '',
+                $unit
+            );
+            take @data.join(';');
+        }
```
jnthn, line 92: I'm still a bit confused about this; `$_` is not used in the loop either, so it will just output the very same data every iteration of the loop?

lib/Agrammon/OutputFormatter/JSON.pm6
```diff
@@ -63,19 +66,17 @@ sub get-data($model, $outputs, $include-filters, $language?, $prints?) {
     return @records.sort(+*.<order>);
 }
 
-sub make-record($fq-name, $output, $model, $raw-value, $var, $order, $instance-id?, :$language, :$prints, :@filters) {
+sub make-record($fq-name, $output, $model, $raw-value, $var, $order, $instance-id?, :$language, :@filters, :@print-set) {
+    my $var-print = $model.output-print($fq-name, $output);
+    next unless $model.should-print($fq-name, $output, @print-set);
```
jnthn, line 71: Probably more efficient to swap these lines around?

lib/Agrammon/OutputFormatter/Text.pm6
```diff
@@ -1,73 +1,79 @@
...
+                            push @output-lines, "        $output = $val";
+                        }
+                        if $include-filters {
+                            if $value ~~ Agrammon::Outputs::FilterGroupCollection && $value.has-filters {
+                                add-filters(@output-lines, $value, $unit, $indent, :$all-filters);
                             }
                         }
                     }
+                    NEXT {
+                        @lines.append: |@title-lines, |@subtitle-lines, |@output-lines if @output-lines;
```
jnthn, line 66: I personally find this more understandable than the `$n` thing. :+1:

lib/Agrammon/OutputFormatter/JSON.pm6
zaucker: Grr, I thought I had caught all of them ... fixed.

lib/Agrammon/OutputFormatter/Text.pm6
zaucker: Then I probably will also when I look at it in the future. I must admit that I had to think about the `$n` version a while also to come up with the explanation. :-)

lib/Agrammon/OutputFormatter/CSV.pm6
zaucker: I removed it because it gave a newline too many because we output with `say`.

lib/Agrammon/Model.pm6
zaucker: Removed, still producing the right output.

lib/Agrammon/OutputFormatter/CSV.pm6
zaucker: Hmm, it does the right thing, because `@filters` only has one element (we don't use multiple filter groups right now).
But you're quite right, I think I fixed it appropriately for now.

It will need some change if we ever use multiple filter groups, so I added a warning about that.

lib/Agrammon/OutputFormatter/CSV.pm6
jnthn: Aha, makes sense.

jnthn (approved):

PR #489, opened 2021-06-22
Add input template REST route

zaucker:
Add a route to get a CSV formatted template of all model inputs with a demo instance for each multi-module.

### DIFF
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -375,7 +375,7 @@ class Agrammon::Model {
     #| string names of their taxonomies. A multi-instance module root is a pair,
     #| where the key is the taxonomy name of the root of the multi-instance module
     #| and the value is an array of the modules within that instance.
-    method extract-structure {
+    method extract-structure() {
         my @singles;
         my %multis;
         for @!evaluation-order -> Agrammon::Model::Module $module {
@@ -516,5 +516,4 @@ class Agrammon::Model {
         };
         %$!distribution-map-cache
     }
-
 }
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -377,17 +377,16 @@ class Agrammon::Model {
     #| and the value is an array of the modules within that instance.
     method extract-structure($sort? = 'calculation') {
         my \order = $sort eq 'calculation' ?? @!evaluation-order.reverse !! @!load-order;
-        my @singles;
-        my %multis;
+        my @modules;
         for order -> Agrammon::Model::Module $module {
             if $module.instance-root -> $root {
-                %multis{$root}.push($module.taxonomy);
+                @modules.push( $root => $module.taxonomy );
             }
             else {
-                @singles.push($module.taxonomy);
+                @modules.push($module.taxonomy);
             }
         }
-        [flat @singles, %multis.pairs]
+        @modules
     }
 
     method run(Agrammon::Inputs :$input!, :%technical --> Agrammon::Outputs) {
--- a/lib/Agrammon/Model.pm6
+++ b/lib/Agrammon/Model.pm6
@@ -378,7 +378,7 @@ class Agrammon::Model {
     method extract-structure {
         my @singles;
         my %multis;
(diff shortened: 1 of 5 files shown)


lib/Agrammon/Model.pm6
```diff
@@ -375,10 +375,10 @@ class Agrammon::Model {
     #| string names of their taxonomies. A multi-instance module root is a pair,
     #| where the key is the taxonomy name of the root of the multi-instance module
     #| and the value is an array of the modules within that instance.
-    method extract-structure() {
+    method extract-structure {
         my @singles;
         my %multis;
-        for @!evaluation-order -> Agrammon::Model::Module $module {
+        for @!evaluation-order.reverse -> Agrammon::Model::Module $module {
```
jnthn, line 381: This bit probably needn't change if it's not being used for the inputs?

lib/Agrammon/Model.pm6
```diff
@@ -389,6 +389,25 @@ class Agrammon::Model {
...
+    #| instance modules. The list returned contains single-instance modules as
+    #| string names of their taxonomies. A multi-instance module root is a pair,
+    #| where the key is the taxonomy name of the root of the multi-instance module
+    #| and the value is an array of the modules within that instance.
+    method extract-structure-sorted($sort? = 'model') {
+        my \order = $sort eq 'calculation' ?? @!evaluation-order !! @!load-order;
+        my @modules;
+        for order -> Agrammon::Model::Module $module {
+            if $module.instance-root -> $root {
+                @modules.push( $root => $module.taxonomy );
```
jnthn, line 402: This will produce loads of pairs rather than a pair mapping to the an array of the modules. That doesn't feel terribly useful; it's creating an intermediate data structure that's not making the next step any easier...

lib/Agrammon/Model.pm6
```diff
@@ -516,4 +535,45 @@ class Agrammon::Model {
         };
         %$!distribution-map-cache
     }
+
+    #| Get a CSV formatted template of all model inputs
+    #| with one demo instance for all multi-instance modules.
+    method get-input-template($sort) {
+        # Obtain model structure in specified order.
+        my @model-structure := self.extract-structure-sorted($sort);
```
jnthn, line 543: Here we could instead do `my @modules := $sort eq 'model' ?? @!load-order !! @!evaluation-order`, and...

lib/Agrammon/Model.pm6
```diff
@@ -516,4 +535,45 @@ class Agrammon::Model {
...
+            when Pair {
+                # Multi-instance module
+                my @modules = @(.value);
+                my $instance-name = 'Demo';
+                for @modules -> $module-name is copy {
+                    my $module = self.get-module($module-name);
+                    add-module-inputs(@inputs, $module, :$instance-name);
+                }
+            }
+        }
```
jnthn, line 561: This lot can be structured as:

```
for @modules -> $module {
    if $module.instance-root -> $root {
        # Multi-instance module
        ...
    }
    else {
        # Single-instance module
        ...
    }
}
```

Note that since you don't have strings but module objects, there's no need to redo the module lookup either.

lib/Agrammon/Model.pm6
```diff
@@ -516,4 +535,45 @@ class Agrammon::Model {
         };
         %$!distribution-map-cache
     }
+
+    #| Get a CSV formatted template of all model inputs
+    #| with one demo instance for all multi-instance modules.
+    method get-input-template($sort) {
```
jnthn, line 541: I don't really think CSV formatting belongs in `Agrammon::Model`; perhaps move this logic out to the service class?

lib/Agrammon/Model.pm6
zaucker: Reverted

lib/Agrammon/Web/Service.pm6
```diff
@@ -205,6 +205,41 @@ class Agrammon::Web::Service {
         };
     }
 
+    # used in method below below
```
jnthn, line 208: 2 x below

jnthn (approved):

PR #491, opened 2021-06-23
Fix JSON output for REST

zaucker:
I guess JSON cannot be streamed?

### DIFF
--- a/lib/Agrammon/Web/APIRoutes.pm6
+++ b/lib/Agrammon/Web/APIRoutes.pm6
@@ -80,9 +80,14 @@ sub rest-api-routes (Str $schema, Agrammon::Web::Service $ws) is export {
                         :language(~$language), :format(~$accept), :print-only(~$print-only),
                         :include-filters($include-filters eq 'true'), :all-filters($all-filters eq 'true')
                     );
-                    content ~$accept, supply {
-                        emit $results.encode('utf8');
-                    };
+                    if ~$accept eq 'application/json' {
+                        content ~$accept, @($results);
+                    }
+                    else {
+                        content ~$accept, supply {
+                            emit $results.encode('utf8');
+                        };
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


lib/Agrammon/Web/APIRoutes.pm6
```diff
@@ -80,9 +80,14 @@ sub rest-api-routes (Str $schema, Agrammon::Web::Service $ws) is export {
...
-                    content ~$accept, supply {
-                        emit $results.encode('utf8');
-                    };
+                    if ~$accept eq 'application/json' {
+                        content ~$accept, @($results);
+                    }
+                    else {
+                        content ~$accept, supply {
+                            emit $results.encode('utf8');
+                        };
```
jnthn, line 89: Hm, since this is only doing one `emit` and the calculation work was already done, this isn't adding anything over just passing the text directly to `content`. For the `supply` approach to really be useful we need to do multiple `emit`s as we produce the data, in order to stream it. That means `get-outputs-from-cs` would also need to produce output in pieces.

I figured the case for streaming output would be when we have many datasets and we `emit` the results for each one. However, it seems here we're only doing a single dataset?

lib/Agrammon/Web/APIRoutes.pm6
```diff
@@ -80,9 +80,14 @@ sub rest-api-routes (Str $schema, Agrammon::Web::Service $ws) is export {
                         :language(~$language), :format(~$accept), :print-only(~$print-only),
                         :include-filters($include-filters eq 'true'), :all-filters($all-filters eq 'true')
                     );
-                    content ~$accept, supply {
-                        emit $results.encode('utf8');
-                    };
+                    if ~$accept eq 'application/json' {
```
jnthn, line 83: `$accept` header will be a `Str` and doesn't need coercing.

jnthn:
> I guess JSON cannot be streamed?

Not exactly but...

1. JSONLines can be (not really applicable here)
2. If you are producing many datasets of output, nothing stops you sending a `[`, then sending the JSON encoding of each data set result, taking care to put in `,`s between, and then a final `]` at the end. It's a bit DIY (you'd have to call `to-json` yourself, etc.)

Although see the comment in the review on why we're not really usefully doing any streaming at the moment.

lib/Agrammon/Web/APIRoutes.pm6
jnthn: Can still drop the `~` here too

jnthn (approved):

lib/Agrammon/Web/APIRoutes.pm6
zaucker: Overlooked and now dropped.

jnthn (approved):

PR #496, opened 2021-06-25
Fix path handling from command line

zaucker:
Handle absolute path for technical file and fix relative path for it from command line

### DIFF
--- a/lib/Agrammon/TechnicalParser.pm6
+++ b/lib/Agrammon/TechnicalParser.pm6
@@ -31,7 +31,14 @@ sub parse-technical(Str $to-parse) is export {
 }
 
 sub load-technical(IO::Path $model-path, Str $technical-file) is export {
-    my $tech-input = $model-path.IO.&child-secure($technical-file // 'technical.cfg');
+    my $tech-input;
+    # handle absolut path
+    if $technical-file and $technical-file ~~ / ^ '/' / {
+        $tech-input = $technical-file;
+    }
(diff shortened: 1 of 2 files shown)


lib/Agrammon/TechnicalParser.pm6
```diff
@@ -31,7 +31,14 @@ sub parse-technical(Str $to-parse) is export {
 }
 
 sub load-technical(IO::Path $model-path, Str $technical-file) is export {
-    my $tech-input = $model-path.IO.&child-secure($technical-file // 'technical.cfg');
+    my $tech-input;
+    # handle absolut path
```
jnthn, line 35: *absolute

lib/Agrammon/TechnicalParser.pm6
```diff
@@ -31,7 +31,14 @@ sub parse-technical(Str $to-parse) is export {
...
+    if $technical-file and $technical-file ~~ / ^ '/' / {
+        $tech-input = $technical-file;
+    }
```
jnthn, line 38: Hm, but this means somebody can submit an absolute path via the web API also and bypass the `child-secure` below?

zaucker: You're right. As I don't see an easy way to handle absolute paths, I just reverted this change, so that I can at least merge the remaining fix.

jnthn (approved):

PR #497, opened 2021-06-27
Handle REST input in JSON format

zaucker:
- Allow input in CSV or JSON (JSON input files being around %43 in size as CSV)
- script for CSV to JSON conversion
- errors and output in JSON or text


lib/Agrammon/DataSource/JSON.pm6
```diff
@@ -0,0 +1,54 @@
...
+            if $full-tax.index('[') -> $sub-start {
+                my $tax = $full-tax.substr(0, $sub-start);
+                my $sub-end = $full-tax.index(']');
+                my $sub-tax = $full-tax.substr($sub-end + 1);
+                $sub-tax = $sub-tax ?? $sub-tax.substr(2) !! '';
+                for $module-data.kv -> $instance, $instance-inputs {
+                    for $instance-inputs.kv -> $var, $value {
+                        $inputs.add-multi-input(
+                                $tax, $instance, $sub-tax,
+                                $var, maybe-numify($value)
```
jnthn, line 21: Do we need to `maybe-numify` in the JSON? Since it has a concept of strings vs. numbers, perhaps we just expect that numeric values are property provided that way?

lib/Agrammon/DataSource/JSON.pm6
```diff
@@ -0,0 +1,54 @@
...
+
+use Agrammon::Inputs;
+use Agrammon::DataSource::Util;
+
+class Agrammon::DataSource::JSON {
+    method load($simulation-name, $dataset-id, $json-data) {
+        my %input-data = from-json $json-data;
+        my $inputs = Agrammon::Inputs.new(:$simulation-name, :$dataset-id);
+        for %input-data.kv -> $full-tax, $module-data {
+            if $full-tax.index('[') -> $sub-start {
```
jnthn, line 12: Not sure how much choice we get about the JSON representation, but maybe it would be better to arrange it so the instance is explicitly given in the JSON rather than having to extract it?

lib/Agrammon/DataSource/JSON.pm6
```diff
@@ -0,0 +1,54 @@
...
+                my $tax = $full-tax.substr(0, $sub-start);
+                my $sub-end = $full-tax.index(']');
+                my $sub-tax = $full-tax.substr($sub-end + 1);
```
jnthn, line 15: Do we want better error handling if the `]` is not found? I think this will probably explode with an exception on the addition.

lib/Agrammon/Web/Service.pm6
```diff
@@ -271,14 +272,27 @@ class Agrammon::Web::Service {
...
     ) {
-        my $input = Agrammon::DataSource::CSV.new.from-csv($simulation-name, $dataset-name, $csv-data);
+        my $data-source;
+        given $type {
+            when 'text/csv' {
+                $data-source = Agrammon::DataSource::CSV.new;
+            }
+            when 'application/json' {
+                $data-source = Agrammon::DataSource::JSON.new;
+            }
```
jnthn, line 293: Maybe a `default` just in case something else slips in?

lib/Agrammon/Web/Service.pm6
```diff
@@ -271,14 +272,27 @@ class Agrammon::Web::Service {
...
         :$include-filters, :$all-filters
     ) {
-        my $input = Agrammon::DataSource::CSV.new.from-csv($simulation-name, $dataset-name, $csv-data);
+        my $data-source;
+        given $type {
+            when 'text/csv' {
+                $data-source = Agrammon::DataSource::CSV.new;
+            }
+            when 'application/json' {
+                $data-source = Agrammon::DataSource::JSON.new;
```
jnthn, line 292: Note that instead of declaring a variable and then assigning in every branch, one can do `my $data-source = do given $type { when 'text/csv' { Agrammon::DataSource::CSV.new } ... }`.

lib/Agrammon/DataSource/JSON.pm6
zaucker: We can define it as we want.
The idea was, to minimize the amount of data to transfer.
What structure would you recommend?

lib/Agrammon/DataSource/JSON.pm6
zaucker: Removed.

lib/Agrammon/DataSource/JSON.pm6
zaucker: I just realized I can search for `[]` in one go, as the instance name is not inside the parens as it is in CSV. Changed.

lib/Agrammon/Web/Service.pm6
zaucker: `$type` is typed as `subset InputFormats` not allowing anything else in. And the OpenAPI definition doesn't allow anything else either. Would you still recommend a default?

lib/Agrammon/Web/Service.pm6
zaucker: Changed.

lib/Agrammon/DataSource/JSON.pm6
jnthn: I wondered about something like:

* For single-instance modules, the top-level keys maps to a hash with the input values
* For multi-instance modules, the part of the taxonomy before the instance maps to an array, which carries an object per instance, with keys `name` (for the instance name) and values, which is then an object mapping the part of the taxonomy after the instance name to an object the the input values

So that we never repeat the instance name or the part of the taxonomy that comes before it.

lib/Agrammon/Web/Service.pm6
jnthn: I was just thinking a simple `die` (doesn't even have to be typed) just in case. But yeah, I agree it can't be reached at present.

zaucker: Ok, added a `die` just in case somebody drops the checks :-)

test/rest/csv2json.pl6
```diff
@@ -0,0 +1,24 @@
...
+    my @rows = $filename.IO.slurp.split("\n");
+
+    my %inputs;
+
+    my ($simulation, $run, $module, $var, $value);
+    for @rows -> $row {
+        next unless $row;
+        ($simulation, $run, $module, $var, $value) = $row.split(';');
+        if $module ~~ / $<root> = [ .+ ] '[' $<instance> = [ .+ ] ']' $<sub> = [ .* ] / {
+            %inputs{~$<root> ~ '[]' ~ (~$<sub> // '')}{~$<instance>}{$var} = $value;
```
jnthn, line 16: I think this needs updating for the new format?

test/rest/json2inputs.pl6
```diff
@@ -0,0 +1,30 @@
+use v6;
+
+use JSON::Fast;
+
+sub MAIN(Str $filename) {
+
+    my %input-data = from-json($filename.IO.slurp)<inputs>;
+    for %input-data.kv -> $full-tax, $module-data {
+        note $full-tax;
+        if $full-tax.index('[') -> $sub-start {
```
jnthn, line 10: This also seems to use the old JSON format.

lib/Agrammon/DataSource/JSON.pm6
```diff
@@ -0,0 +1,165 @@
...
+                                $full-tax, $instance, $sub-tax,
+                                $var, $value
+                            );
+                        }
+                    }
+                }
+            }
+            else {
+                for $module-data -> $input-hash {
+                    for $input-hash.kv -> $var, $value {
```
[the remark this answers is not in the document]
zaucker: Did this.

lib/Agrammon/DataSource/JSON.pm6
```diff
@@ -0,0 +1,165 @@
...
+                            $inputs.add-multi-input(
+                                $full-tax, $instance, $sub-tax,
+                                $var, $value
+                            );
+                        }
+                    }
+                }
+            }
+            else {
+                for $module-data -> $input-hash {
```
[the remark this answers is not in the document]
zaucker: Used fix below

lib/Agrammon/DataSource/JSON.pm6
```diff
@@ -0,0 +1,165 @@
...
+
+class Agrammon::DataSource::JSON {
+    method load($simulation-name, $dataset-id, $json-data) {
+        my %input-data = from-json $json-data;
+        my $inputs = Agrammon::Inputs.new(:$simulation-name, :$dataset-id);
+        for %input-data.kv -> $full-tax, $module-data {
+            if $module-data ~~ Array {
+                for @($module-data) {
+                    my $instance = $_<name>;
+                    my $values = $_<values>;
```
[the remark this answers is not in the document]
zaucker: Done

test/rest/csv2json.pl6
zaucker: Deleted.

test/rest/json2inputs.pl6
zaucker: Deleted (together with some other test scripts and test data files.

lib/Agrammon/Web/Service.pm6
```diff
@@ -271,14 +272,23 @@ class Agrammon::Web::Service {
         $model-path.IO.&child-secure($technical).slurp
     }
 
-    method get-outputs-from-csv(
-        Agrammon::DB::User $user,
-        Str $simulation-name, Str $dataset-name, $csv-data,
+    subset InputFormats of Str where  { $_ eq 'application/json' or $_ eq 'text/csv' };
+    subset OutputFormats of Str where { $_ eq 'application/json' or $_ eq 'text/csv' or $_ eq 'text/plain' };
```
jnthn, line 276: Missed this before, but: since in non-block form a `subset`'s `where` part is smart-matched against the value, you can write these was:

```
    subset InputFormats of Str where 'application/json' | 'text/csv';
    subset OutputFormats of Str where 'application/json' | 'text/csv' | 'text/plain';
```

t/datasource-json.t
```diff
@@ -0,0 +1,38 @@
...
+
+my $filename = 't/test-data/inputs-version6.json';
+
+my $inputs;
+subtest "Load inputs from JSON" => {
+    ok my $input-data = $filename.IO.slurp, "Read JSON file";
+    isa-ok my $data-source = Agrammon::DataSource::JSON.new, Agrammon::DataSource::JSON,
+    'Is a Agrammon::DataSource::JSON';
+    ok $inputs = $data-source.load('Agrammon Test', 'Version6', $input-data), "Load inputs";
+    note $inputs.all-inputs.keys.join("\n");
```
jnthn, line 16: Debugging leftover, I think

jnthn (approved):
One probably leftover debug thing that should go, and one potential cleanup I guess I missed before, but there's nothing much wrong now.

lib/Agrammon/Web/Service.pm6
zaucker: Nice. Done.

t/datasource-json.t
zaucker: Removed.

PR #498, opened 2021-06-27
Make command line arguments more consistent

zaucker:


lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -28,13 +28,14 @@ my %*SUB-MAIN-OPTS =
   :named-anywhere,    # allow named variables at any location
 ;
 
+subset OptionalExistingFile of Str where { .IO.e or note("No such file $_") && exit 1 }
```
jnthn, line 31: This has been added, but isn't being used

lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -28,13 +28,14 @@ my %*SUB-MAIN-OPTS =
...
 subset ExistingFile of Str where { .IO.e or $_ eq '-' or note("No such file $_") && exit 1 }
 subset SupportedLanguage of Str where { $_ ~~ /^ de|en|fr $/ or note("ERROR: --language=[de|en|fr]") && exit 1 };
 subset SortOrder of Str where { $_ ~~ /^ model|calculation $/ or note("ERROR: --sort=[model|calculation]") && exit 1 };
 subset OutputFormat of Str where { $_ ~~ /^ csv|json|text $/ or note("ERROR: --format=[csv|json|text]") && exit 1 };
 
 #| Start the web interface
-multi sub MAIN('web', ExistingFile $cfg-filename, Str $model-filename, Str $technical-file?) is export {
+multi sub MAIN('web', Str $model-filename, Str :$cfg-filename, Str :$technical-file) is export {
```
jnthn, line 38: I'm curious why the dropping of the subset type?

lib/Agrammon/UI/CommandLine.pm6
zaucker: Ah, yes, decided I can't use it ... see below. Removed.

lib/Agrammon/UI/CommandLine.pm6
zaucker: Because I thought I couldn't restrict it being now optional. I add the type for `:$cfg-filename` again.
I think I cannot restrict `$technical-file` or `$model-filename`, though, as those are being `add()` ed to the base path.
Rename `$cfg-filename` to `$cfg-file` to be consistent with `$technical-file` (and it is less to type).

jnthn: Trouble is that:

```
subset OptionalExistingFile of Str where { $_ ~~ Any or .IO.e or note("No such file $_") && exit 1 }
```

Will always match because just about anything matches against `Any`. I think you want is:

```
subset OptionalExistingFile of Str where { !.defined or .IO.e or note("No such file $_") && exit 1 }
```

Which will permit an undefined value, but if it's defined it must be a file that exists.

zaucker: Indeed, that's what I wanted :-) ... changed.

lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -28,14 +28,15 @@ my %*SUB-MAIN-OPTS =
   :named-anywhere,    # allow named variables at any location
 ;
 
-subset ExistingFile of Str where { .IO.e or $_ eq '-' or note("No such file $_") && exit 1 }
+subset OptionalExistingFile of Str where { !.defined or .IO.e or note("No such file $_") && exit 1 }
+subset ExistingFile of Str where { $_ ~~ Any or .IO.e or $_ eq '-' or note("No such file $_") && exit 1 }
```
jnthn, line 32: I think it wants to be changed here too, or if it's not optional then the `$_ ~~ Any` just removing?

jnthn (approved):

PR #511, opened 2021-08-30
Allow flattened/branched sums to be not exactly 100%

zaucker:


jnthn (commented):
Two thoughts:

1. Should it not be `abs($sum - 100)` in case it's a tiny bit wrong in the other direction?
2. Since we do the same check twice, factor it out to a sub?

(Also, fwiw, [there is an operator](https://docs.raku.org/routine/=~=) for "roughly equal", but its default tolerance is more aimed at floating point errors.)

jnthn (approved):

PR #512, opened 2021-09-01
Make printf output more robust

zaucker:
The old code blew up with `$title = 'foo 100% bar'`

The not so helpful error message was:
```
Your printf-style directives specify 2 arguments, but 1 argument was supplied.
Are you using an interpolated '$'?
[ERROR] 500 /get_output_variables - 10.34.12.100
```
More helpful would be if the format string was part of the output.

### DIFF
--- a/lib/Agrammon/Performance.pm6
+++ b/lib/Agrammon/Performance.pm6
@@ -4,6 +4,6 @@ sub timed(Str $title, &proc) is export {
     my $start = now;
     my \ret   = proc;
     my $end   = now;
-    note sprintf "$title ran %.3f seconds", $end-$start;
+    note sprintf "%s ran %.3f seconds", $title, $end-$start;
     return ret;
 }


jnthn (approved):

jnthn:
I guess I missed this in code review; in general, it's not a good idea to mix interpolation and `sprintf`, because one can end up with meta-characters in the string, and you've discovered, and...

> Are you using an interpolated '$'?

...I think that is exactly what this bit of the error message was trying to tell you about.

> More helpful would be if the format string was part of the output.

Yes, given it should usually not be too long.

Also this is a good idea for a Comma annotation to avoid making this mistake in the first place; I'll make an issue for us to work on that.

zaucker:
Yes, the meaning of the error message is clear enough and having encountered it, I think I'll remember not to do this again.
In a code base with lots of printf statements it would help to get the culprit shown as well (assuming that filename/line number aren't available at runtime).

Comma helper is certainly a good idea.

zaucker:
> I'm a bit confused why you don't have a backtrace; are you catching and printing the exception with just the message?
```
                content 'application/json', { %output };
                CATCH {
                    default {
                        note "$_";
                        response.status = 500;
                        content 'application/json', %( error => .message )
```
I guess I should rather `dd` instead of `note`?

jnthn:
I think `note $_` (also spelled `.note`) would do it; the problem is that the quotes call `.Str` and an exception stringifies to just its message (while `note` given it directly will `.gist` it, which includes the trace).

zaucker:
> I think `note $_` (also spelled `.note`) would do it; the problem is that the quotes call `.Str` and an exception stringifies to just its message (while `note` given it directly will `.gist` it, which includes the trace).

Oh, shorted code, better outcome! Will go through `Routes.pm` and make a PR for it.

PR #516, opened 2021-09-07
Add standard value to dimensioning_barn

ChHaeni:


t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock/DairyCow/Excretion.nhd
```diff
@@ -87,26 +87,29 @@ Flisch R, Sinaj S, Charles R, Richner W 2009. Grundlagen für die die Düngung i
        <p>Nombre effectif d'animaux dans la stabulation.</p>
 
 +dimensioning_barn
-  type = float
+  type = text
   distribute = true
+  default_calc = Standard
+  default_gui  = Standard
+  default_formula = In(animals);
```
jnthn, line 94: I think `default_calc` and `default_formula` should be mutually exclusive (and in the future we should start to detect that); thus the `default_calc` should removed here (`default_gui` should stay along with `default_formula`, however).

zaucker:
I wonder if the output dimensioning_barn is needed.

default_formula on the input should make calculations with an undefined input just work.

ChHaeni:
> I wonder if the output dimensioning_barn is needed.
> 
> default_formula on the input should make calculations with an undefined input just work.

Yes it is not needed, but one could possibly detect errors if included in the output..?

t/test-data/Models/hr-inclNOxExtendedWithFilters/Livestock/DairyCow/Excretion.nhd
ChHaeni: Has been removed in c1a8c29.

zaucker:
> > I wonder if the output dimensioning_barn is needed.
> > default_formula on the input should make calculations with an undefined input just work.
> 
> Yes it is not needed, but one could possibly detect errors if included in the output..?

Please call to discuss this. The current formula is not correct for the way `default_formula` works.

zaucker (approved):
I think the output `dimensioning_barn` would not really be necessary unless used in some other module. Instead, the input could be used directly inside the module.
I will merge this nevertheless for the `default_formulas`. We can discuss this after my vacation.

PR #520, opened 2021-09-11
Add getDataset command

zaucker:
Implement export of datasets from DB as CSV

### DIFF
--- a/lib/Agrammon/UI/CommandLine.pm6
+++ b/lib/Agrammon/UI/CommandLine.pm6
@@ -7,12 +7,14 @@ use DB::Pg;
 use JSON::Fast;
 
 use Agrammon::Config;
+use Agrammon::DataSource::DB;
 use Agrammon::DataSource::CSV;
 use Agrammon::Documentation;
 use Agrammon::ModelCache;
 use Agrammon::OutputFormatter::CSV;
 use Agrammon::OutputFormatter::JSON;
 use Agrammon::OutputFormatter::Text;
+use Agrammon::OutputFormatter::Util;
 use Agrammon::Performance;
 use Agrammon::ResultCollector;
 use Agrammon::TechnicalParser;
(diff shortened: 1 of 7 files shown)


lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -95,6 +97,44 @@ multi sub MAIN('run', Str $filename, ExistingFileOrStdin $input, ExistingFile :$
...
+multi sub MAIN(
+        'getDataset', Str $model-filename, Str $username, Str $dataset-name,
+        Str $version, Str $gui, Str $variant, ExistingFile :$cfg-file
+    ) is export {
+    my %data := get-dataset $model-filename, $username, $dataset-name, $version, $gui, $variant, :$cfg-file;
+    my @output;
+    @output.push("## User: %data<username>");
+    @output.push("## Dataset: %data<dataset>");
+    if %data<validation-errors> {
+        note "WARNING: validation errors";
```
jnthn, line 110: Why do we `note` this rather than putting it into the output?

lib/Agrammon/UI/CommandLine.pm6
```diff
@@ -95,6 +97,44 @@ multi sub MAIN('run', Str $filename, ExistingFileOrStdin $input, ExistingFile :$
...
+    for $inputs.kv -> $module, $_ {
+        when Hash {
+            for sorted-kv($_) -> $input, $value {
+                @output.push("$module;$input;$value");
+            }
+        }
+        when Array {
+            for sorted-kv($_) -> $instance-id, %instance-inputs {
+                for sorted-kv(%instance-inputs) -> $fq-name, %values {
+                    my $q-name = $module ~ '[' ~ $instance-id ~ ']' ~ $fq-name.substr($module.chars);
```
jnthn, line 127: I suspect we do this calculation (of the name from module, instance ID, and fully qualified name) in enough places while producing output that it may just be worth factoring out into ` Agrammon::OutputFormatter::Util`.

lib/Agrammon/UI/CommandLine.pm6
zaucker: > Why do we `note` this rather than putting it into the output?

Because I expect the user to redirect the output to a file and possibly overlook the fact that are validation errors.
The next lines are writing the actual errors to  the output.

jnthn: Ah, OK, that makes sense.

lib/Agrammon/UI/CommandLine.pm6
zaucker: Done

jnthn (approved):

zaucker:
Force merging as GHA is broken (Ghostscript install).

PR #521, opened 2021-09-12
FixPDF Output

zaucker:
- Handle slash in dataset name for filename
- Extend log collector
-  Fix log output

### DIFF
--- a/lib/Agrammon/OutputFormatter/PDF.pm6
+++ b/lib/Agrammon/OutputFormatter/PDF.pm6
@@ -36,8 +36,8 @@ sub create-pdf($temp-dir-name, $pdf-prog, $timeout, $username, $dataset-name, %d
     }
 
     my $filename = "agrammon_export_" ~ $username ~ "_$dataset-name";
-    # cannot have / in filenames
-    $filename ~~ s:g/ \/ / - /;
+    # sanitize internally used filename
+    $filename ~~ s:g/<-[\w\s_-]>/-/;
     my $source-file = "$temp-dir/$filename.tex".IO;
     my $pdf-file    = "$temp-dir/$filename.pdf".IO;
     my $aux-file    = "$temp-dir/$filename.aux".IO;
--- a/lib/Agrammon/OutputFormatter/PDF.pm6
+++ b/lib/Agrammon/OutputFormatter/PDF.pm6
@@ -36,6 +36,8 @@ sub create-pdf($temp-dir-name, $pdf-prog, $timeout, $username, $dataset-name, %d
     }
 
     my $filename = "agrammon_export_" ~ $username ~ "_$dataset-name";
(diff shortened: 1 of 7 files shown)


lib/Agrammon/OutputFormatter/PDF.pm6
```diff
@@ -36,6 +36,8 @@ sub create-pdf($temp-dir-name, $pdf-prog, $timeout, $username, $dataset-name, %d
     }
 
     my $filename = "agrammon_export_" ~ $username ~ "_$dataset-name";
+    # cannot have / in filenames
+    $filename ~~ s:g/ \/ / - /;
```
jnthn, line 40: Wonder if it's worth having an allow list instead (e.g. `s:g/<-[\w\s_-]>/-/` or so), given exactly what different OSes permit is quite variable...

lib/Agrammon/OutputFormatter/PDF.pm6
```diff
@@ -162,6 +164,17 @@ sub input-output-as-pdf(
...
+    %data<log> = @log;
+    my @entries := $outputs.log-collector.entries;
+    for @entries -> $_ {
+        @log.push(%(
+            message  => latex-escape(.messages{$language}),
+            gui      => .gui{$language},
+            instance => latex-escape(.instance),
+        ));
+    }
+
```
jnthn, line 177: I see `@log` being set up, but then not used anywhere

zaucker: > I see `@log` being set up, but then not used anywhere

It is used on the template:
```
<@log>
...
</@>
```

lib/Agrammon/OutputFormatter/PDF.pm6
zaucker: > Wonder if it's worth having an allow list instead (e.g. `s:g/<-[\w\s_-]>/-/` or so), given exactly what different OSes permit is quite variable...

I figured this was good enough, as the server is running on Linux only (at least for now), but will implement your  suggestion for easier extension in the future.
The name is actually used only for creation of the PDF code on the server and not passed to the user or her OS.

zaucker (commented):
Not sure why I ended up with a review comment ...

zaucker:
The checks fail because GHA suddenly cannot install LaTeX for some reason (actually the Ghostscript library) :-(

lib/Agrammon/OutputFormatter/PDF.pm6
jnthn: Oh, I totally missed the line `%data<log> = @log;` - probably because I tend to set things up and *then* stash them where they are going, so expected it to be at the end. :-)

zaucker:
Force merging while GHA is broken with Ghostscript installation.

PR #525, opened 2021-09-30
Improve defaults handling in GUI

zaucker:
- Leave null values in inputs when sending to GUI (was forced to empty string before)
- Show `Standard` in GUI if input value is null and has a default
- Count as "complete" in this case
- Some cleanup of table cell renderers
- Write input validation errors to server log


frontend/source/translation/fr.po
```diff
@@ -974,3 +974,7 @@ msgstr ""
 #: agrammon/Changelog.js
 msgid "Changelog"
 msgstr ""
+
+#. NO LONGER USED
+msgid "Standard"
+msgstr ""
```
jnthn, line 980: If it's no longer used, it can't be removed?

lib/Agrammon/DataSource/Util.pm6
```diff
@@ -1,5 +1,6 @@
 use v6;
 
 sub maybe-numify($value) is export {
+    return $value if not $value.defined;
```
jnthn, line 4: Or `.return without $value;`.

lib/Agrammon/Web/Service.pm6
```diff
@@ -350,6 +350,7 @@ class Agrammon::Web::Service {
         my $results = self!get-outputs($user, $dataset-name)<results>;
         my $validation-errors = self!get-outputs($user, $dataset-name)<validation-errors>;
         warn '**** Got ' ~  $validation-errors.elems ~ ' input validation errors' if $validation-errors;
+        dd 'Validation errors:', $validation-errors if $validation-errors;
```
jnthn, line 353: Debug leftover?

frontend/source/translation/fr.po
zaucker: removed

lib/Agrammon/DataSource/Util.pm6
zaucker: Neat. Done.

lib/Agrammon/Web/Service.pm6
zaucker: No, on purpose for the moment. I think it is useful to have those in the log file on the server (see also comment on this PR at the top).
Added a comment.

jnthn (approved):

PR #526, opened 2021-09-30
Hide upload button from regular users and better error messages

zaucker:
- Upload format isn't documented yet and thus just confuses regular users
- Handle upload errors better


zaucker:
Small frontend change, merging right away

PR #530, opened 2021-10-22
Allow none-integer distributable values

zaucker:
In the regional model, the customer wants to allow none-integer "number of animals or staple places".
This was implemented in the model already, but apparently not used so far and currently fails as `dist-total` is enforced to be `Int`. Change to `Numeric`.


jnthn (approved):
"Now with support for fractional cows!"

PR #533, opened 2021-12-09
Demo datasets

zaucker:
- Add demo dataset info to frontend
- Partially cleanup some frontend code


zaucker:
Merging for release. @jnthn Please have a look at the Raku part of this part whenever you have time. Only a few minor changes which I can fix easily if something is wrong with them.

PR #535, opened 2021-12-10
Handle expiration

zaucker:
- Set session expiration to 2 hours
- Fix relogin after expiration
- Some frontend cleanup


zaucker:
Merging, frontend only (except for longer session duration which is a trivial backend change).

PR #536, opened 2021-12-11
Move models

zaucker:
Move current model to `share/Models/version6`  and symlink to test-data directory


zaucker:
Just moved model around. Merging.

PR #537, opened 2021-12-11
Frontend improvements

zaucker:
- Change main menue from Toolbar to Menubar
- Handle `Standard` as special acceptable value for number types

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -8058,9 +8058,9 @@
       }
     },
     "node_modules/caniuse-lite": {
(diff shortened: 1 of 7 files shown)


zaucker:
@jnthn Please have a look at the change in `Validation.pm6`. Unfortunately, we cannot get rid of this work-around completely as discussed in the chat. And this PRs allows to have the correct type for numeric inputs anyway instead of the previously used `text`  type that didn't validate numbers at all.

zaucker:
Merging for new release.

PR #539, opened 2022-01-10
RT53007: remove default_calc on selectBox

zaucker:
`default_calc = none` on the selectBox for `mitigation_housing_floor` effectively forced this value on flattened instances overriding the real value.

**NOTE: ** `default_calc` must not be used on SelectBoxes with more than one item.


zaucker:
Model fix, merging

PR #556, opened 2022-06-08
Clean Up code related to default captures

ChHaeni:


ChHaeni:
Sorry, accidently pushed commit for next PR on top of this one

zaucker (commented):
Did you test that removing the checks for definedness don't lead to problems if an input file is used with missing values? E.g., are such errors caught by the input validation?

ChHaeni:
> Did you test that removing the checks for definedness don't lead to problems if an input file is used with missing values? E.g., are such errors caught by the input validation?

Not specifically. However, I think it's better if the model crashes, rather than silently proceeding with some default value which is not appropriate. 

Which input variables are validated - do they need a specific validator entry or are e.g. enums validated, too?

I'm intending to add some explicit checks on some input variables, where the model tells the user about the invalid input and crashes afterwards - unless there is a throw/exception like option?

zaucker (approved):

PR #563, opened 2022-06-16
Feature to be merged towards end of the year

ChHaeni:
Split effect of floor mitigation to `indoor` and `outdoor`, thus, allowing to define different indoor/outdoor reductions depending on selected floor mitigation.


ChHaeni:
Sorry that was too quick...

PR #578, opened 2023-04-11
Extend preprocessor logic

zaucker:
- add ?or and ?and to ?if clause
- add tests
- add tests for unsupported usage

### DIFF
--- a/lib/Agrammon/Preprocessor.pm6
+++ b/lib/Agrammon/Preprocessor.pm6
@@ -51,16 +51,18 @@ class X::Agrammon::Preprocessor::ElsifMultiOptions does X::Agrammon::Preprocesso
 #| Preprocess the provided source code for C-style preprocessor directives (but
 #| with a different syntax, given `#` is taken as the comment character). The
 #| syntax available is:
-#|     ?if FOO
+#|     ?if FOO and FOO2 and FOO3 ...
 #|     foo stuff
-#|     ?elsif BAR
+#|     ?elsif BAR or BAR2 or BAR3 ...
 #|     bar stuff
 #|     ?else
 #|     other stuff
 #|     ?endif
-#| Where the C<elsif> and C<else> parts are optional. The symbols FOO and BAR
-#| will be looked up and truth-tested in C<%options>. Nesting is allowed. An
-#| error will be raised on mis-nesting.
+#| Where the C<elsif> and C<else> parts are optional. The symbols FOO, FOO2,
+#| BAR, and BAR2 will be looked up and truth-tested in C<%options>.
+#| and/or parts are optional, but must not be mixed on one clause.
+#| Nesting is allowed. An error will be raised on mis-nesting.
+
 sub preprocess(Str $source, %options --> Str) is export {
     my @result-lines;
     my class OpenDirective {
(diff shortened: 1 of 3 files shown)


jnthn (commented):
This isn't a terribly elegant solution; there's:

* Reparsing (the split, the re-identitying the negation)
* Code duplication between the cases
* A bit too much strictness on whitespace around the operators (hard-coded spaces instead of `\h+`)
* Potential bugs because `.+` in Raku includes newlines
* Fairly arbitrary of exclusion of `elsif` support for the feature

In my opinion the `?or` is also overkill, since the `?` already said we're in a preproc directive, so I think it can be just `or` there.

I think it would overall be much cleaner to write one rule for all the and/or/single cases like this:

```
when /^ '?if' $<parts>=(\h+ [$<negate>='!']? <option=.ident>)+ % [\h+ $<op>=< and or >] \h* $/ {
}
```

That gets you a neat, structured, match tree. While we could go and do proper precedence, let's for now accept only `or`s or only `and`s; we can detect the mixing with:

```
unless ‌‌all([eq] $<op>) {
    die X::Agrammon::Preprocessor::MixedAndOr.new: :line($number + 1);
}
```

From there it depends if a short-circuit is desired or not. Let's assume not. Then we can do:

```
my @enabled-parts = $<parts>.map: -> $/ {
     $<negate> ?? .not !! .so given %options{~$<option>}
}
```

And then collapse them to a single boolean value:

```
my $enabled = $<op> && $<op>[0] eq 'or'
    ?? ?any(@enabled-parts)
    !! ?all(@enabled-parts);
```

Noting that the `all` handles the case of no operator, because `?all($bool)` is the same as `$bool`.

Given how much less additions this causes, it's probably now not so terrible to implement it for `elsif` too (especially as the processing can be extracted to a `sub`).

zaucker:
> This isn't a terribly elegant solution; there's:

Re-implemented it with your suggestions, thanks!

lib/Agrammon/Preprocessor.pm6
```diff
@@ -64,18 +80,18 @@ sub preprocess(Str $source, %options --> Str) is export {
...
                     @open.push(OpenDirective.new(:$start-line, :$enabled, :$matched, :accept-elsif));
                 }
-                when /^ '?elsif' \h+ [$<negate>='!']? <option=.ident> \h* $/ {
+                when /^ '?elsif' $<parts>=(\h+ [$<negate>='!']? <option=.ident>)+ % [\h+ $<op>=< and or >] \h* $/ {
                     if @open.pop -> $prev-part {
                         unless $prev-part.accept-elsif {
                             die X::Agrammon::Preprocessor::ElsifAfterElse.new: :line($number + 1);
                         }
-                        my $enabled = so !$prev-part.matched && ($<negate> ?? .not !! .so given %options{~$<option>});
+                        my $enabled = so !$prev-part.matched && process-match($<parts>, $<negate>, %options, $<op>, $<option>, $number);
```
jnthn, line 94: It's a nitpick but the `so` is not needed here given `!` and `process-match` both return booleans.

jnthn (commented):
About fine now, one bit of dead code to remove.

lib/Agrammon/Preprocessor.pm6
```diff
@@ -34,19 +34,35 @@ class X::Agrammon::Preprocessor::ElsifAfterElse does X::Agrammon::Preprocessor {
...
+        "Invalid preprocessor ?and and ?or in same ?if found at line $!line"
+    }
+}
+
+#| Exception thrown when ?elsif with multiple options
+class X::Agrammon::Preprocessor::ElsifMultiOptions does X::Agrammon::Preprocessor {
+    method message() {
+        "Invalid preprocessor ?elsif with multiple options found at line $!line"
+    }
+}
```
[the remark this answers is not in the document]
zaucker: Removed.

lib/Agrammon/Preprocessor.pm6
zaucker: Removed.

zaucker:
One question: why is the `|$op` needed here
```
sub process-match($parts, $negate, %options, $op, $option, $number --> Bool) {
    unless all([eq] |$op) {
        die X::Agrammon::Preprocessor::MixedAndOr.new: :line($number + 1);
    }
```
while it is not needed if the test ist done inside `preprocess()` like this
```
    unless all([eq] |$<op>) {
```
And how can I tell the difference between `$<op>` in the caller and `$op` in the called sub?

jnthn:
`$<op>` is syntax sugar for `$/<op>`, and what you get back is a `List`, which `[eq]` then iterates.

Since `$` means item, then when receiving it into `$op` it gets wrapped in a `Scalar` container, and then `[eq]` sees it as a single item.

You might prefer to use `@op` in the signature instead, since `$<op>` gives a `List` of matches back, so that's the better sigil for it.

jnthn (approved):

zaucker:
Thanks for the explanation. Now I understand the `$[]` versus `[]` which `dd` showed (wrapping in Scalar container).

I guess I should use `@` for the other lists as well.

Will update the PR.
> $<op> is syntax sugar for $/<op>, and what you get back is a List, which [eq] then iterates.
> 
> Since $ means item, then when receiving it into $op it gets wrapped in a Scalar container, and then [eq] sees it as a single item.
> 
> You might prefer to use @op in the signature instead, since $<op> gives a List of matches back, so that's the better sigil for it.

zaucker:
I realized that I hadn't quite understood the nesting of the regex match and thus passed useless (`Nil`) parameters to `process-match()`. Removed them and used `@` for the lists being passed.

jnthn (approved):

PR #590, opened 2023-11-22
Hide hiddenAnimals input

zaucker:
Not sure where this is used, but it should not be part of models used in production.


zaucker:
This was already fixed it seems

PR #620, opened 2026-01-06
Implement `for` loops and array literals

jnthn:
For loops mostly use the Raku syntax (`->` style), however since our `if` and `given` all currently require parens around the condition or topic, we require them here too for consistency. We allow multiple loop variables for iterating two items at a time.

For hashes, we do a key/value iteration, so you iterate the two at once and put them into separate variables. This isn't quite Perl or Raku, but it is the path of least resistance from what we have today to getting something useful. (We also don't go down the path of sigils beyond the existing `$`.)

Finally, array literals like `[1, 2]` are also introduced. They can be used directly with `for` or assigned to a variable. There are no other operations on arrays for now; they are introduced primarily in support of `for`.

An example just like the request in https://github.com/oposs/agrammon/issues/600 is included among the tests, to make sure we can handle that use case.

### DIFF
--- a/lib/Agrammon/Environment.rakumod
+++ b/lib/Agrammon/Environment.rakumod
@@ -20,4 +20,8 @@ class Agrammon::Environment {
     method find-builtin($name) {
         %!builtins{$name} // get-builtins(){$name} // die "No such builtin function '$name'";
     }
+
+    method iterate($value) {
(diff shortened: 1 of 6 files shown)


zaucker (approved):
I'll merge and have asked Christoph to have a look if he sees any additional features that might be useful.

PR #629, opened 2026-05-18
Add model version switcher under Options menu

zaucker:
## Summary

Adds a model version switcher driven by a new optional `Versions:` block in the deployment YAML. Companion to the cross-version enum-alias support already on `main` (merged via #628) — together they let users navigate between sibling deployments of different model versions without losing access to their existing datasets.

**Configuration**
- New `Versions:` block in deployment YAML — list of `{ label, url, version, title }` entries pointing at sibling instances. Leave the block out (or empty) to hide the switcher entirely.
- `Agrammon::Config.versions` accessor + `get_cfg` exposes the list to the SPA.

**Frontend**
- New "Set model version ..." submenu under **Options**, sibling to "Set language ...". The entry matching the active version is disabled so you can't switch to yourself. Strictly-older picks trigger a Confirm dialog.
- Active-version `title:` from the matching `Versions` entry overrides `GUI.title` in the page header so users always see which version they're in.
- Dataset table now lists datasets across sibling model versions and fades rows whose `dataset_version` doesn't match the active `Model.version` (new `agrammon.ui.table.rowrenderer.DatasetVersion`). The version column is made visible so the mismatch is legible.

**Backend**
- `DB::Datasets.load` no longer filters by `dataset_version`, so foreign-version datasets show up (and the frontend fades them).

**Session sharing**
- Already on `main` (resume_session route + SPA call): same Postgres-backed session cookie is honored across sibling instances, so switching versions does NOT prompt for re-login.

## Test plan

- [x] `prove6 -l t/config.rakutest` — new positive test for `Versions:` parsing (2 new subtests, fixture YAML)
- [x] `prove6 -l t/webservice.rakutest` — `get_cfg` includes `versions => []` when no block configured
- [ ] `make test` — full suite against test DB
- [ ] Manual: configure two sibling instances on different ports, log in on one, pick the other in **Options → Set model version ...** — header title updates, no re-login, dataset table shows datasets from both versions with the non-matching one faded
- [ ] Manual: pick a strictly-older version — Confirm dialog appears before navigation

### DIFF
--- a/etc/agrammon.yaml.dist
+++ b/etc/agrammon.yaml.dist
@@ -26,3 +26,31 @@ Model:
     technical: technical.cfg
     variant:   SHL
     version:   "6.0 - #REV#"
+
+# Sibling model versions reachable from this deployment.
+# Each entry: { label, url, version, title }
+#   - label    short string shown in the dropdown
+#   - url      where to navigate when this version is picked
+#   - version  matches the Model.version of the sibling instance;
+#              the entry whose version equals THIS process's
+#              Model.version is the active one. Switching to a strictly
+#              older version triggers a confirm dialog in the GUI.
+#   - title    per-locale title shown in the GUI header for THIS version.
+#              The active entry's title overrides GUI.title.
+# Order should be newest first.
+# Leave the block out (or empty) to hide the dropdown.
+Versions:
+    - label:   '7.0'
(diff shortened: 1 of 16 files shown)


zaucker:
Merged after visual local test

PR #637, opened 2026-06-02
perf(outputs): share singleton empty FilterKey (~4%); repair model-cache test

zaucker:
## Summary

Three small, low-risk commits split out from a larger working branch:

1. **`perf(outputs)`: share a singleton empty `FilterKey` in `FilterGroupCollection`** — every scalar value flowing through the model is wrapped into a single-group `FilterGroupCollection` keyed by an *empty* `FilterKey`. A warm v7.0.0 / HR_2015 run made ~1973 `from-scalar` calls, each allocating a fresh empty `FilterKey` and recomputing an identical, immutable `WHICH`. `FilterKey` is a value object compared purely by `WHICH` with a never-mutated `%.filters`, so a single shared empty instance is semantically identical. **~4% faster per dataset** (median 177.4ms → 170.7ms, paired, 100 reps).
2. **`test(model-cache)`: repair the model-cache test** — the test loaded the cached model into a random `$TMPDIR` never placed on the module search path, so every cached load died ("Failed 5/6"). Put the real cache dir (`~/.agrammon`) on the path with `use lib` before any Agrammon module — the same contract `bin/agrammon.raku` uses to dodge the Rakudo 2026.04 compunit-identity drift. Cache is content-hashed, so sharing the production dir is safe.
3. **`Update docu`** — comment-only clarifications in `XLSXWriter.rakumod`'s header (no logic change).

## Risk

Low. The optimization is **output-neutral** — verified by comparing the full output-value multiset before vs after (identical). The test fix is test-only; the docs commit is comment-only.

## Verification

- Full unit suite: **49 files / 529 tests, all PASS** (`AGRAMMON_UNIT_TEST=1 prove6 -l t/`) — up from 48/49 before the model-cache repair.
- `output-filter-group-collection.rakutest` 56/56 (covers the FilterKey path).
- Branch is current with `main` (0 behind) — clean fast-forward.

### DIFF
--- a/lib/Agrammon/OutputFormatter/XLSXWriter.rakumod
+++ b/lib/Agrammon/OutputFormatter/XLSXWriter.rakumod
@@ -4,17 +4,18 @@ use Libarchive::Simple;
 # A small, fast, write-only XLSX generator.
 #
 # It builds the worksheet XML as strings (no DOM, no per-node FFI) with a shared
-# style table referenced by index — the way Excel itself and Excel::Writer::XLSX
+# style table referenced by index — the way Excel itself and Perl5's Excel::Writer::XLSX
 # store workbooks — then packs the parts into the .xlsx zip with Libarchive (the
 # same library Spreadsheet::XLSX already uses for zipping).
 #
 # Why this exists: the DOM-based Spreadsheet::XLSX path is ~30x slower than the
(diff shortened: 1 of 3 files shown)


zaucker:
Merging as all tests are ok

PR #660, opened 2026-06-18
feat(#569): parallel batch runs over the REST API

### LINKED ISSUE #569: Extend REST API to include parallelization
Up to now, it is no yet possible to run the model in parallel via REST interface.

zaucker:
Closes #569. Implements the approach from the issue assessment comment.

## What
Extends the `runSimulation` REST operation so one request can run **many datasets in parallel**. When `simulation` **and** `dataset` are both omitted, the payload is treated as a multi-dataset batch: every dataset runs concurrently against one shared, read-only model, and the response is an array of per-dataset results. Supplying `simulation`+`dataset` keeps the existing single-dataset behaviour byte-for-byte unchanged.

### Batch payload formats
- **CSV** — the existing multi-dataset layout `simulation;dataset;taxonomy;variable;value` (new sync `DataSource::CSV.read-data`).
- **JSON** — an array of `{ simulation, dataset, data }` objects (new `DataSource::JSON.load-data`); single-dataset `load` refactored to share the builder.

### Concurrency
`get-batch-outputs-for-rest` runs datasets via the CLI's proven `race … .race(:degree)` pattern. `degree` is an optional, **bounded (1..16, default 4)** knob so a single request can't starve the server.

### Per-dataset isolation
A dataset that fails validation yields `{ simulation, dataset, validationErrors }`; one that dies during the run yields `{ …, error }`. One bad dataset never aborts the batch.

## Concurrency-safety fix (important)
The shared model + output machinery build several lazy `||=` caches (output labels/format/order, …) on first use, and that build is **not atomic**. Running datasets in parallel from cold raced and produced wrong — sometimes mismatched — results (reproduced at ~60% failure under stress). The batch now runs the **first dataset single-threaded to warm all shared state** (its result is kept, not wasted) before racing the rest. Verified deterministic across heavy repetition.

## Route / OpenAPI
- `simulation`/`dataset` now optional; new `degree` field; batch JSON returns an `outputs` array; CSV/text concatenate per dataset (failures as comment lines); Excel output is rejected for batch with a clear 400.
- `share/agrammon-rest.openapi` updated accordingly.

## Tests
- New `t/run-batch-rest.rakutest`: JSON & CSV batch, **batch == single-run parity**, per-dataset error isolation, and a **12-dataset concurrency stress** subtest (degree 8) asserting all results are identical (regression guard for the race).
- Datasource batch parsers covered in `datasource-{csv,json}`; route batch dispatch in `apiroutes`; integration test in `webservice`.
- **Full unit suite green: 568 tests / 52 files.** (Pre-existing `webservice.rakutest` DB-state failures on `rename-dataset` are unrelated — confirmed identical on clean `main`.)

## Benchmark (12 datasets, 128-core host, median of 5 reps)
| degree | median ms | speedup |
|--------|-----------|---------|
| 1 | 1686 | 1.00× |
| 2 | 1006 | 1.68× |
| 4 | 625 | 2.70× |
| 8 | 504 | 3.34× |

Diminishing returns are expected: the single-threaded warm-up dataset is a fixed serial fraction (Amdahl).

### DIFF
--- a/.github/workflows/backend-and-models.yml
+++ b/.github/workflows/backend-and-models.yml
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   raku:
-    timeout-minutes: 30
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
--- a/lib/Agrammon/DataSource/CSV.rakumod
+++ b/lib/Agrammon/DataSource/CSV.rakumod
@@ -26,23 +26,49 @@ class Agrammon::DataSource::CSV {
     }
 
     method !group-input(@group) {
-        start {
-            my $input = Agrammon::Inputs.new(simulation-name => @group[0][0], dataset-id => @group[0][1]);
-            for @group {
-                my $full-tax = .[2];
-                if $full-tax.index('[') -> $sub-start {
-                    my $tax = $full-tax.substr(0, $sub-start);
-                    my $sub-end = $full-tax.index(']');
-                    my $instance = $full-tax.substr($sub-start + 1, ($sub-end - $sub-start) - 1);
-                    my $sub-tax = $full-tax.substr($sub-end + 1);
-                    $input.add-multi-input($tax, $instance, $sub-tax ?? $sub-tax.substr(2) !! '', .[3], maybe-numify(.[4]))
-                }
-                else {
-                    $input.add-single-input(.[2], .[3], maybe-numify(.[4]));
-                }
+        start { self!build-input(@group) }
+    }
+
+    #| Build one Agrammon::Inputs from a group of rows in the multi-dataset CSV
+    #| layout (columns: simulation; dataset; taxonomy; variable; value).
+    method !build-input(@group) {
+        my $input = Agrammon::Inputs.new(simulation-name => @group[0][0], dataset-id => @group[0][1]);
+        for @group {
+            my $full-tax = .[2];
+            if $full-tax.index('[') -> $sub-start {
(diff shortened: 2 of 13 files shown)


zaucker:
## Benchmark: parallel speedup by `degree`

Workload: **12 datasets** (identical copies of `t/test-data/inputs-version6.json`) run through `get-batch-outputs-for-rest` with `format=application/json`, `print-only=SummaryTotal`. Host: 128-core Linux, Rakudo 2026.04. Median of 5 reps after one untimed warm-up call (so cold JIT/spesh + model-cache costs don't skew `degree=1`).

| degree | median ms | min ms | speedup vs degree=1 |
|-------:|----------:|-------:|--------------------:|
| 1 | 1686 | 1632 | 1.00× |
| 2 | 1006 | 1003 | 1.68× |
| 4 |  625 |  619 | 2.70× |
| 8 |  504 |  492 | 3.34× |

Every rep was checked for correctness (all 12 totals identical, no race).

**Why it plateaus (Amdahl):** the batch deliberately runs the **first dataset single-threaded** to warm the model's non-atomic lazy `||=` output caches before parallelizing — its result is kept, not wasted, but it's a fixed serial fraction (~1 of 12 dataset-runs, ≈140 ms here). At infinite `degree` the floor is roughly `warm-up + one parallel batch` ≈ 2 dataset-times, so ~3–4× is about what an 8-way split of 12 small datasets yields. Larger batches amortize the serial warm-up better and scale further.

